PR #2, opened 2021-06-07
[PAG-287] infra app service mockPa

pasqualespica:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/app-service.tf
+++ b/src/app-service.tf
@@ -1,0 +1,39 @@
+resource "azurerm_resource_group" "example" {
+  name     = format("%s-law", local.project)
+  location = var.location
+}
+
+resource "azurerm_app_service_plan" "app_service_plan" {
+  name                = var.plan_name
+  location            = var.location
+  resource_group_name = var.resource_group_name
+
+  sku {
+    tier = "Standard"
+    size = "S1"
+  }
+}
+
+resource "azurerm_app_service" "app_service" {
+  name                = var.name
+  location            = var.location
+  resource_group_name = var.resource_group_name
+  app_service_plan_id = azurerm_app_service_plan.app_service_plan.id
+  https_only          = true
+  client_cert_enabled = true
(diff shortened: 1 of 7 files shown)


src/monitor.tf
```diff
@@ -0,0 +1,21 @@
+resource "azurerm_resource_group" "monitor_rg" {
```
pasqualedevita, line 1: you can copy from here
https://github.com/pagopa/cstar-infrastructure/blob/main/src/monitor.tf

src/app-service.tf
```diff
@@ -0,0 +1,39 @@
+resource "azurerm_resource_group" "example" {
+  name     = format("%s-law", local.project)
+  location = var.location
+}
```
pasqualedevita, line 4:
```suggestion
resource "azurerm_resource_group" "mockec_rg" {
  name     = format("%s-mockec-rg", local.project)
  location = var.location
}
```

src/app-service.tf
```diff
@@ -0,0 +1,39 @@
...
+
+resource "azurerm_app_service_plan" "app_service_plan" {
```
pasqualedevita, line 6: maybe it's better to create an app service module
https://github.com/pagopa/cgn-onboarding-portal-infra/blob/master/code/api.tf#L35

pasqualedevita (changes requested):

src/main.tf
```diff
@@ -11,7 +11,7 @@ terraform {
     hostname     = "app.terraform.io"
     organization = "PagoPa"
     workspaces {
-      prefix = "prefix_template-"
+      prefix = "pagopa-"
```
GiovanniMancini, line 14: @pasqualespica Maybe e less general prefix name would be a better fit here

PR #21, opened 2021-09-28
[PCFG-9] api-config app service

jacopocarlini:
### List of changes

- added new app service for api-config
- added new api product api-config
- added new api for api-config product

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualedevita (changes requested):
1. to use a link to a openapi/swagger definition maybe it's better to point to a github raw url
2. using `10.1.240.16/29` we have a few ips to scale out this service (maybe it isn't a requirement)

pasqualespica:
> 1. to use a link to a openapi/swagger definition maybe it's better to point to a github raw url (using app service link we have a chicken eggs problem :) )
> 2. using `10.1.240.16/29` we have a few ips to scale out this service (maybe it isn't a requirement)

they were both my suggestions 
1. the first one was to point out automatically to openspec definition. @jacopocarlini we use as suggested by @pasqualedevita raw file on github
2. it is exactly not a requirement ( no scale out, only internal use )

pasqualedevita (changes requested):
api-config should be inside vnet_integration

pasqualespica:
> api-config should be inside vnet_integration



> api-config should be inside vnet_integration

right, modified

pasqualespica (approved):

pasqualedevita (approved):

pasqualedevita (approved):

PR #26, opened 2021-09-29
temp disable redis cache apim

pasqualespica:
Temporary disabling of the redis cache for apim

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualedevita (commented):
a better approach is to create a new var `redis_cache_enabled` and using count.
In this way terraform code it's ready to add redis cache

pasqualedevita (approved):

pasqualedevita:
top @pasqualespica

PR #56, opened 2021-11-05
Change function to use app service plan

pasqualedevita:
### List of changes

use app service plan for azure functions
need this PR https://github.com/pagopa/azurerm/pull/111

APPLY IN PROD WILL DESTROY THE FUNCTION

### Motivation and context

cost savings

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/checkout_function.tf
+++ b/src/checkout_function.tf
@@ -37,7 +37,9 @@ module "checkout_function" {
   health_check_path                        = "api/v1/info"
   subnet_out_id                            = module.checkout_function_snet[0].id
   runtime_version                          = "~3"
+  always_on                                = var.checkout_function_always_on
   application_insights_instrumentation_key = azurerm_application_insights.application_insights.instrumentation_key
+
 
   app_service_plan_info = {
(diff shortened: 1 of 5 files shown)


src/checkout_function.tf
```diff
@@ -76,3 +79,66 @@ module "checkout_function" {
...
+    name = "default"
+
+    capacity {
+      default = var.checkout_function_autoscale_default
+      minimum = var.checkout_function_autoscale_minimum
+      maximum = var.checkout_function_autoscale_maximum
+    }
+
+    rule {
+      metric_trigger {
```
infantesimone, line 101: the idea is to scale only according to the number of requests received in a period of time? if so I think it's fine, maybe with some performance tests I'm curious to understand the correlation with the CPU. (I wouldn't really take memory into account, due to the code of these API)

infantesimone (approved):
waiting for https://github.com/pagopa/azurerm/pull/111

infantesimone (approved):

PR #58, opened 2021-11-08
Add load balancer for db nodo SIA

pasqualespica:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/lb.tf
+++ b/src/lb.tf
@@ -1,0 +1,66 @@
+module "lb_snet" {
+  source               = "git::https://github.com/pagopa/azurerm.git//subnet?ref=v1.0.51"
+  name                 = format("%s-lb-snet", local.project)
+  resource_group_name  = azurerm_resource_group.rg_vnet.name
+  virtual_network_name = module.vnet_integration.name
+  address_prefixes     = var.cidr_subnet_lb
+
+  enforce_private_link_endpoint_network_policies = true
+  service_endpoints                              = ["Microsoft.Web"]
+}
+
+
+module "integration_lb" {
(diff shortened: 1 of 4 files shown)


src/lb.tf
```diff
@@ -0,0 +1,66 @@
...
+          ip      = "10.101.35.40"
+          vnet_id = module.vnet_integration.id
+        },
+        {
+          ip      = "10.101.35.41"
+          vnet_id = module.vnet_integration.id
+        }
+      ]
+    }
+  ]
```
pasqualedevita, line 45: better to use a var

src/lb.tf
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  lb_probe = {
+    tcp-8022 = {
+      protocol     = "Tcp"
+      port         = "8022"
+      request_path = ""
+    }
+  }
```
pasqualedevita, line 63: check these ports for oracledb

pasqualedevita (changes requested):

src/variables.tf
```diff
@@ -555,3 +555,20 @@ variable "acr_enabled" {
...
+  type        = list(string)
+  description = "Address prefixes load balancer"
+  default     = null
+}
+variable "lb_integration_frontend_ip" {
+  type    = string
+  default = null
+}
+
+variable "lb_enabled" {
```
uolter, line 570: Is there any environment where the load balancer is not needed?
I mean: do we really need this flag?

uolter (changes requested):

uolter (dismissed):

PR #61, opened 2021-11-09
[PCFG-84] api versioning

jacopocarlini:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/apim_api_config.tf
```diff
@@ -35,6 +45,9 @@ module "apim_api_config_api" {
   product_ids           = [module.apim_api_config_product[0].product_id]
   subscription_required = true
 
+  version_set_id = azurerm_api_management_api_version_set.api_config_api[0].id
+  api_version    = "v1"
```
pasqualespica, line 49: remember this change it'll introcude something like that on url request path `https://api.*.platform.pagopa.it/apiconfig/api/v1/<endpoint>`

pasqualespica (approved):
👍

pasqualedevita (approved):

PR #71, opened 2021-11-17
[IOPAY-346] alerts for monitoring

infantesimone:
### List of changes

- add `azurerm_application_insights_web_test`to monitor frontend;
- add `azurerm_monitor_scheduled_query_rules_alert` to monitor checkout availability;
- add `azurerm_monitor_metric_alert` to monitor 5xx of checkout function app;

### Motivation and context

The goal of this PR is to create some alarms for the checkout API.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/checkout_fe.tf
+++ b/src/checkout_fe.tf
@@ -88,7 +88,7 @@ resource "azurerm_application_insights_web_test" "checkout_fe_web_test" {
   count                   = var.checkout_enabled ? 1 : 0
   name                    = format("%s-checkout-fe-web-test", local.project)
   location                = var.location
-  resource_group_name     = azurerm_resource_group.checkout_fe_rg[0].name
+  resource_group_name     = azurerm_resource_group.monitor_rg.name
   application_insights_id = azurerm_application_insights.application_insights.id
   kind                    = "ping"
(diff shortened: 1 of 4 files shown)


src/checkout_fe.tf
```diff
@@ -83,3 +83,26 @@ module "checkout_cdn" {
 
   tags = var.tags
 }
+
+resource "azurerm_application_insights_web_test" "checkout_fe_web_test" {
```
pasqualedevita, line 87: since we added health check path in function u can use `Health Check Status` metrics, in this way u'll ping all function instances

infantesimone: This test is not for the function app. The goal of this test is to monitor if the front-end exposes resources, that is, if the GET dns/index.html responds 200.

pasqualedevita (dismissed):

src/checkout_fe.tf
pasqualedevita: thanks, my mistake I saw function

pasqualedevita (approved):

PR #73, opened 2021-11-18
[PCFG-94] Routing

aomegax:
Added dynamic rule regarding `url_file_extension_condition` in order to manage routing through React and avoid to find resource in the Azure Storage

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (approved):
well done 💪

pasqualedevita (approved):

PR #89, opened 2021-12-27
[IOPAY-403] Add infra to support `pagopa-proxy`

giovanniberti:
### List of changes

This PR adds the necessary infrastructure to support a (updated) copy of `pagopa-proxy`, which is currently under the IO subscription.

### Motivation and context

This PR is part of the ongoing process of migrating infrastructure necessary for communicating with "Nodo" from the IO subscription to PagoPA subscription (during the migration phase there will be two concurrently running versions of the proxy, one for each subscription).

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/env/dev/terraform.tfvars
```diff
@@ -23,6 +23,7 @@ cidr_subnet_mock_psp = ["10.1.240.8/29"]
 
 cidr_subnet_checkout_be = ["10.1.240.17/29"]
 cidr_subnet_buyerbanks  = ["10.1.240.24/29"]
+cidr_subnet_proxy       = ["10.1.240.32/29"]
```
infantesimone, line 26: `proxy` too generic, better `pagopaproxy` (using the same naming convention on the resources present in[ IO subscription](https://github.com/pagopa/io-infrastructure-live-new/blob/43d89ce34ecef945dadefdb1dc6dc5473a1bd20a/prod/westeurope/external/pagopaproxyprod/app_service/terragrunt.hcl#L52) ). 
p.s it is also valid for all resources/variables of this PR

src/env/dev/terraform.tfvars
```diff
@@ -232,3 +233,11 @@ buyerbanks_function_autoscale_minimum = 1
 buyerbanks_function_autoscale_maximum = 3
 buyerbanks_function_autoscale_default = 1
 buyerbanks_delete_retention_days      = 30
+
+# proxy app service
+proxy_enabled        = true
+proxy_redis_capacity = 1
+proxy_redis_sku_name = "Standard"
```
infantesimone, line 240: For DEV env, to change sku name in "Basic" --> https://azure.microsoft.com/en-au/pricing/details/cache/

p.s for example --> https://github.com/pagopa/selfcare-infra/blob/439a5a34bca237095eb7a58cb670eff03a41c5ab/src/core/env/dev/terraform.tfvars#L44

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,146 @@
...
+    service_delegation = {
+      name    = "Microsoft.Web/serverFarms"
+      actions = ["Microsoft.Network/virtualNetworks/subnets/action"]
+    }
+  }
+}
+
+module "proxy_redis" {
+  count                 = var.proxy_enabled ? 1 : 0
+  source                = "git::https://github.com/pagopa/azurerm.git//redis_cache?ref=v1.0.37"
```
infantesimone, line 30: why not use the latest version? 2.0.19?

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,146 @@
...
+  name                  = format("%s-proxy-redis", local.project)
+  resource_group_name   = azurerm_resource_group.proxy_rg[0].name
+  location              = azurerm_resource_group.proxy_rg[0].location
+  capacity              = var.proxy_redis_capacity
+  enable_non_ssl_port   = true
+  family                = var.proxy_redis_family
+  sku_name              = var.proxy_redis_sku_name
+  enable_authentication = true
+  subnet_id             = length(module.proxy_snet[0].*.id) == 0 ? null : module.proxy_snet[0].id
```
infantesimone, line 39: why not only module.proxy_snet[0].id?

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,146 @@
...
+  plan_reserved = true # Mandatory for Linux plan
+
+  # App service plan
+  name                = format("%s-app-proxy", local.project)
+  client_cert_enabled = false
+  always_on           = var.proxy_always_on
+  linux_fx_version    = "NODE|14-lts"
+  health_check_path   = "/ping"
+
+  app_settings = {
```
infantesimone, line 65: To review app settings according to the current configuration for IO subscription --> https://github.com/pagopa/io-infrastructure-live-new/blob/43d89ce34ecef945dadefdb1dc6dc5473a1bd20a/prod/westeurope/external/pagopaproxyprod/app_service/terragrunt.hcl#L70

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,146 @@
...
+    InstrumentationEngine_EXTENSION_VERSION         = "disabled"
+    SnapshotDebugger_EXTENSION_VERSION              = "disabled"
+    XDT_MicrosoftApplicationInsights_BaseExtensions = "disabled"
+    XDT_MicrosoftApplicationInsights_Mode           = "recommended"
+    XDT_MicrosoftApplicationInsights_PreemptSdk     = "disabled"
+    WEBSITE_HEALTHCHECK_MAXPINGFAILURES             = 10
+    TIMEOUT_DELAY                                   = 300
+
+    # Proxy-specific env vars
+    # Taken from https://github.com/pagopa/io-infrastructure/blob/6f43d141e7695b96cdb192535a30b78efdd9df81/infrastructure/kubernetes/pagopa-proxy.yml
```
infantesimone, line 83: see previous comment

infantesimone (changes requested):

giovanniberti:
Should we set up a private endpoint for redis (along with a new dns zone maybe)?

infantesimone:
> Should we set up a private endpoint for redis (along with a new dns zone maybe)?

Good point, yes... Only for dev env we can disable private endpoint, for example https://github.com/pagopa/selfcare-infra/blob/439a5a34bca237095eb7a58cb670eff03a41c5ab/src/core/redis_cache.tf#L21

src/env/dev/terraform.tfvars
```diff
@@ -232,3 +233,12 @@ buyerbanks_function_autoscale_minimum = 1
 buyerbanks_function_autoscale_maximum = 3
 buyerbanks_function_autoscale_default = 1
 buyerbanks_delete_retention_days      = 30
+
+# pagopa-proxy app service
+pagopa_proxy_enabled                        = true
+pagopa_proxy_redis_capacity                 = 1
```
pasqualedevita, line 239:
```suggestion
pagopa_proxy_redis_capacity                 = 0
```
zero is the min value

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,147 @@
...
+  tags = var.tags
+}
+
+module "pagopa_proxy_snet" {
+  count                = var.pagopa_proxy_enabled && var.cidr_subnet_pagopa_proxy != null ? 1 : 0
+  source               = "git::https://github.com/pagopa/azurerm.git//subnet?ref=v2.0.19"
+  name                 = format("%s-pagopa-proxy-snet", local.project)
+  address_prefixes     = var.cidr_subnet_pagopa_proxy
+  resource_group_name  = azurerm_resource_group.rg_vnet.name
+  virtual_network_name = module.vnet.name
```
pasqualedevita, line 15: how pagopa-proxy will contact pagopa-nodo apis?

src/pagopa_proxy.tf
```diff
@@ -0,0 +1,147 @@
...
+  subnet_id             = module.pagopa_proxy_snet[0].id
+
+  tags = var.tags
+
+  private_endpoint = {
+    enabled              = var.pagopa_proxy_redis_private_endpoint_enabled
+    virtual_network_id   = azurerm_resource_group.rg_vnet.id
+    subnet_id            = module.pagopa_proxy_snet[0].id
+    private_dns_zone_ids = var.pagopa_proxy_redis_private_endpoint_enabled ? [azurerm_private_dns_zone.privatelink_redis_pagopa_proxy[0].id] : []
+  }
```
pasqualedevita, line 48: I don't know it we can put in the same subnet pagopa-proxy app service and redis private endpoint

pasqualedevita (changes requested):

src/dns_private.tf
```diff
@@ -22,3 +22,11 @@ resource "azurerm_private_dns_a_record" "private_dns_a_record_db_nodo" {
   ttl                 = 60
   records             = var.dns_a_reconds_dbnodo_ips
 }
+
+resource "azurerm_private_dns_zone" "privatelink_redis_pagopa_proxy" {
```
pasqualedevita, line 26: U have to link private dns zone with virtual network using virtual network link

example:
https://github.com/pagopa/selfcare-infra/blob/main/src/core/dns_private.tf#L64

pasqualedevita (changes requested):

src/dns_private.tf
giovanniberti: Thanks! Fixed in 1f43f52

src/pagopa_proxy.tf
infantesimone: The goal is to use apim pagoPA - SIA connection, so (in another PR) we will map the nodo services on the apim (used from pagopa-proxy)

src/dns_private.tf
```diff
@@ -22,3 +22,22 @@ resource "azurerm_private_dns_a_record" "private_dns_a_record_db_nodo" {
   ttl                 = 60
   records             = var.dns_a_reconds_dbnodo_ips
 }
+
+resource "azurerm_private_dns_zone" "redis_pagopa_proxy_dns_zone" {
+  count               = var.pagopa_proxy_redis_private_endpoint_enabled ? 1 : 0
```
infantesimone, line 27: `var.pagopa_proxy_redis_private_endpoint_enabled  && pagopa_proxy_enabled ? 1 : 0`

infantesimone (changes requested):

src/dns_private.tf
```diff
@@ -22,3 +22,22 @@ resource "azurerm_private_dns_a_record" "private_dns_a_record_db_nodo" {
   ttl                 = 60
   records             = var.dns_a_reconds_dbnodo_ips
 }
+
+resource "azurerm_private_dns_zone" "redis_pagopa_proxy_dns_zone" {
```
pasqualedevita, line 26:
```suggestion
resource "azurerm_private_dns_zone" "privatelink_redis_cache_windows_net" {
```
private dns zone and virtual network links names should be generic: we can have only 1 dns zone and 1 virtual network link for vnet. So if u have 2 redis cache u have to reuse this dns zone and vnet link.
I suggest to use `privatelink_redis_cache_windows_net` and `privatelink_redis_cache_windows_net_vnet` as here https://github.com/pagopa/selfcare-infra/blob/main/src/core/dns_private.tf#L56

src/dns_private.tf
```diff
@@ -22,3 +22,22 @@ resource "azurerm_private_dns_a_record" "private_dns_a_record_db_nodo" {
...
+  count               = var.pagopa_proxy_redis_private_endpoint_enabled ? 1 : 0
+  name                = "privatelink.redis.cache.windows.net"
+  resource_group_name = azurerm_resource_group.data.name
+
+  tags = var.tags
+}
+
+resource "azurerm_private_dns_zone_virtual_network_link" "redis_pagopa_proxy_vnet_link" {
```
pasqualedevita, line 34:
```suggestion
resource "azurerm_private_dns_zone_virtual_network_link" "privatelink_redis_cache_windows_net_vnet" {
```

src/dns_private.tf
pasqualedevita: it's correct to use only `pagopa_proxy_redis_private_endpoint_enabled`

for dns u can use a global var redis_private_endpoint_enabled not related to pagopa_proxy

pasqualedevita (changes requested):

src/dns_private.tf
giovanniberti: What about also renaming `pagopa_proxy_redis_private_endpoint_enabled` to `redis_private_endpoint_enabled`?

If understood everything correctly the virtual network link and the dns zone can be used by many different redis instances (as long as they are under the same vnet), so it should make sense to generify the variable enabling those too.

pasqualedevita: it's correct we will reuse the redis dns zone and vnet link in future redis instances. Now we have two ways:
1. use a single variable `redis_private_endpoint_enabled`
2. set a count condition with OR, for example

```
count = (var.pagopa_proxy_redis_private_endpoint_enabled || var.service_b_redis_private_endpoint_enabled || var.service_c_redis_private_endpoint_enabled) ? 1 : 0
```

PS: we can remove pagopa_proxy_enabled from dns zone and vnet link count condition

infantesimone (approved):

src/env/dev/terraform.tfvars
```diff
@@ -232,3 +238,11 @@ buyerbanks_function_autoscale_minimum = 1
...
 buyerbanks_function_autoscale_default = 1
 buyerbanks_delete_retention_days      = 30
+
+# pagopa-proxy app service
+pagopa_proxy_enabled        = true
+pagopa_proxy_redis_capacity = 0
+pagopa_proxy_redis_sku_name = "Basic"
+pagopa_proxy_redis_family   = "C"
+pagopa_proxy_tier           = "Basic"
+pagopa_proxy_size           = "B1"
```
pasqualedevita, line 248:
```suggestion
pagopa_proxy_tier           = "Standard"
pagopa_proxy_size           = "S1"
```
not blocking but...
basic sku does not support vnet integration required to connect to internal apim vnet

pasqualedevita (approved):

PR #94, opened 2022-01-04
[IOPAY-405] Update APIM config for `pagopa-proxy` 

giovanniberti:
### List of changes

 - Add `pagopa-proxy` APIs to APIM

### Motivation and context

This PR exposes `pagopa-proxy` APIs through APIM.
In this phase both `pagopa-functions-checkout` and `pagopa-proxy` have the capability of communicating with Nodo (the former through the `pagopa-proxy` instance under the IO subcription). The long-term plan will be to make the API exposed through this PR the only communication point towards Nodo (see related RFC).

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/apim.tf
+++ b/src/apim.tf
@@ -88,6 +88,15 @@ resource "azurerm_api_management_named_value" "pagopa_fn_checkout_url_value" {
   value               = format("https://pagopa-%s-fn-checkout.azurewebsites.net", var.env_short)
 }
 
+resource "azurerm_api_management_named_value" "pagopa_appservice_proxy_url_value" {
+  count               = var.checkout_enabled ? 1 : 0
+  name                = "pagopa-appservice-proxy-url"
+  api_management_name = module.apim.name
+  resource_group_name = azurerm_resource_group.rg_api.name
+  display_name        = "pagopa-appservice-proxy-url"
+  value               = format("https://pagopa-%s-app-pagopa-proxy.azurewebsites.net", var.env_short)
+}
+
 resource "azurerm_api_management_named_value" "brokerlist_value" {
   name                = "brokerlist"
   api_management_name = module.apim.name
--- a/src/apim_checkout.tf
+++ b/src/apim_checkout.tf
@@ -3,7 +3,7 @@
 ##############
 
 module "apim_checkout_product" {
-  count  = var.checkout_enabled ? 1 : 0
+  count  = var.checkout_enabled || var.pagopa_proxy_enabled ? 1 : 0
   source = "git::https://github.com/pagopa/azurerm.git//api_management_product?ref=v1.0.16"
 
   product_id   = "checkout"
@@ -29,6 +29,14 @@ locals {
     display_name          = "Checkout payment activation API"
     description           = "API to support payment activation"
     path                  = "api/checkout/payments"
+    subscription_required = false
+    service_url           = null
+  }
+
+  apim_checkout_payment_activations_api = {
+    display_name          = "Checkout payment activations API [new]"
+    description           = "API to support payment activation [new]"
+    path                  = "api/checkout/activations"
     subscription_required = false
     service_url           = null
   }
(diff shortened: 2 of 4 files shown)


src/pagopa_proxy.tf
```diff
@@ -97,7 +97,7 @@ module "pagopa_proxy_app_service" {
     REDIS_DB_URL      = module.pagopa_proxy_redis[0].hostname
     REDIS_DB_PORT     = module.pagopa_proxy_redis[0].ssl_port
     REDIS_DB_PASSWORD = module.pagopa_proxy_redis[0].primary_access_key
-    REDIS_USE_CLUSTER = true
+    REDIS_USE_CLUSTER = var.env_short == "p"
```
infantesimone, line 100: why not create a new variable? related to the env?
for example, we could also decide in UAT to set this parameter to true (maybe 😁)

src/apim_checkout.tf
```diff
@@ -3,7 +3,7 @@
 ##############
 
 module "apim_checkout_product" {
-  count  = var.checkout_enabled ? 1 : 0
+  count  = var.checkout_enabled || var.pagopa_proxy_enabled ? 1 : 0
```
pasqualespica, line 6: why `||` ? 
it shouldn't be `&&` ?

src/apim.tf
```diff
@@ -88,6 +88,15 @@ resource "azurerm_api_management_named_value" "pagopa_fn_checkout_url_value" {
   value               = format("https://pagopa-%s-fn-checkout.azurewebsites.net", var.env_short)
 }
 
+resource "azurerm_api_management_named_value" "pagopa_appservice_proxy_url_value" {
+  count               = var.pagopa_proxy_enabled ? 1 : 0
+  name                = "pagopa-appservice-proxy-url"
+  api_management_name = module.apim.name
+  resource_group_name = azurerm_resource_group.rg_api.name
+  display_name        = "pagopa-appservice-proxy-url"
+  value               = format("https://pagopa-%s-app-pagopa-proxy.azurewebsites.net", var.env_short)
```
pasqualespica, line 97: u could use something like `format("https://%s", module.pagopa_proxy_app_service[0].default_site_hostname)`

src/apim_checkout.tf
```diff
@@ -73,6 +81,113 @@ module "apim_checkout_payments_api_v1" {
...
+  description  = local.apim_cd_info_wisp.description
+  display_name = local.apim_cd_info_wisp.display_name
+  path         = local.apim_cd_info_wisp.path
+  protocols    = ["https"]
+
+  soap_pass_through = true
+
+  import {
+    content_format = "wsdl-link"
+    content_value  = "https://raw.githubusercontent.com/pagopa/io-pagopa-proxy/v0.20.1-RELEASE/src/wsdl/CdPerNodo.wsdl"
```
pasqualespica, line 166: why here reference to wsdl link instead of local file , as other soap import ?

src/apim_nodo_services.tf
```diff
@@ -174,3 +174,174 @@ resource "azurerm_api_management_api" "apim_nodo_pagamenti_api_ec_svr" {
...
+  versioning_scheme   = "Segment"
+}
+
+resource "azurerm_api_management_api" "apim_node_for_psp_api_v1" {
+  name                  = format("%s-node-for-psp-api", var.env_short)
+  api_management_name   = module.apim.name
+  resource_group_name   = azurerm_resource_group.rg_api.name
+  subscription_required = local.apim_node_for_psp_api.subscription_required
+  version_set_id        = azurerm_api_management_api_version_set.node_for_psp_api.id
+  version               = "v1"
```
pasqualespica, line 205: future needs , we should link this version to https://github.com/pagopa/pagopa-api version

pasqualespica (approved):
great work 👍

src/apim_checkout.tf
giovanniberti: It was made more to make the file version explicit than anything else, but I can can include the file directly in the repo to make it consistent with the other wsdl definitions.

src/apim_checkout.tf
giovanniberti: `module.checkout_function` is enabled through `checkout_enabled`, while `pagopa-proxy` infra is enabled through `pagopa_proxy_enabled` and both need on `apim_checkout_product`.

Should the proxy or the checkout function not be enabled this product must be present in order to expose everything to the APIM correctly (at least for how things are configured now).

infantesimone (approved):
🚀

infantesimone (approved):

infantesimone (approved):

pasqualespica (approved):
👍

uolter (approved):

PR #95, opened 2022-01-10
Upd postgresql config

pasqualespica:
Updated postgresql configuration 

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/env/dev/terraform.tfvars
```diff
@@ -85,6 +85,88 @@ postgresql_network_rules = {
 }
 prostgresql_db_mockpsp = "mock-psp"
 
+db_enable_replica = false
+postgresql_configuration = {
```
uolter, line 89: can we move this configurations inside the variables.tf file so that you don't have to make them explicit in every environment. Especially when the is disabled.

uolter (approved):

PR #96, opened 2022-01-10
[PM-316] Update app settings for buyerbank functions

MarcoAGreco:
### List of changes
- Extend app settings for buyerbank functions
- Enable buyerbank functions for UAT

### Motivation and context

- Complete the required setup for buyerbank functions
- Add buyerbanks resources for UAT env

### Type of changes

- [X] Add new resources

- [X] No

- [ ] Yes
- [ ] No

### Other information

---


src/apim_buyerbanks.tf
```diff
@@ -0,0 +1,72 @@
+##############
+## Products ##
+##############
+
+module "apim_buyerbanks_product" {
+  count  = var.buyerbanks_enabled ? 1 : 0
```
uolter, line 6: I think we are going to use too often this count with <something>_enabled
If the resource needs to be created in all environment just create  it.
As far as, I know it's not a best practice to use count everywhere

uolter (changes requested):

uolter (approved):

uolter (approved):

uolter (approved):

src/apim.tf
```diff
@@ -230,6 +230,27 @@ resource "azurerm_api_management_named_value" "checkout_google_recaptcha_secret"
   secret              = true
 }
 
+resource "azurerm_api_management_named_value" "pagopa_fn_buyerbanks_url_value" {
+  count               = var.buyerbanks_enabled ? 1 : 0
```
uolter, line 234: Can we remove the count here?

PR #97, opened 2022-01-11
DB Sit Nodo Pagamenti

jacopocarlini:
### List of changes

- changed DNS name
- updated swagger apiconfig

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/env/dev/terraform.tfvars
+++ b/src/env/dev/terraform.tfvars
@@ -228,7 +228,7 @@ acr_enabled = true
 db_port                            = 1523
 db_service_name                    = "NDPSPCT_PP_NODO4_CFG"
 dns_a_reconds_dbnodo_ips           = ["10.8.3.228"]
-private_dns_zone_db_nodo_pagamenti = "dev.db-nodo-pagamenti.com"
+private_dns_zone_db_nodo_pagamenti = "d.db-nodo-pagamenti.com"
 
 # API Config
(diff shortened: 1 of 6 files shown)


src/env/dev/terraform.tfvars
```diff
@@ -225,9 +225,10 @@ eventhubs = [
 acr_enabled = true
 
 # db nodo dei pagamenti
-db_port                  = 1523
-db_service_name          = "NDPSPCT_NODO4_CFG" # fixme set with data from SIA
-dns_a_reconds_dbnodo_ips = ["10.8.3.228"]
+db_port                            = 1523
+db_service_name                    = "NDPSPCT_PP_NODO4_CFG"
+dns_a_reconds_dbnodo_ips           = ["10.8.3.228"]
+private_dns_zone_db_nodo_pagamenti = "dev.db-nodo-pagamenti.com"
```
pasqualespica, line 231: could be bettere use `env` var instead of _dev_ , _uat_ , _prod_ strings  ?

pasqualespica: info : as suggests from @uolter we can do this, we choose an arbitrary dns name, iif there isn't a TSL connection between SIA DB

pasqualespica (approved):

PR #111, opened 2022-01-18
[PPR-291] Infra for pagopa-nodo-dei-pagamenti-test

aomegax:
Added CDN endpoint and storage account resources.

### List of changes
- added resource configuration for cdn and storage account

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/nodo_dei_pagamenti_test.tf
+++ b/src/nodo_dei_pagamenti_test.tf
@@ -1,0 +1,31 @@
+/**
+ * nodo-dei-pagamenti resource group
+ **/
+resource "azurerm_resource_group" "nodo_pagamenti_test_rg" {
+  count    = var.nodo_pagamenti_test_enabled ? 1 : 0
+  name     = format("%s-nodo-pagamenti-test-rg", local.project)
+  location = var.location
+
+  tags = var.tags
+}
+
+/**
+ * CDN
+ */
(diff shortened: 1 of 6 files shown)


src/nodo_dei_pagamenti_test.tf
```diff
@@ -0,0 +1,33 @@
...
+* STORAGE
+*/
+module "nodo_test_storage" {
+  source = "git::https://github.com/pagopa/azurerm.git//storage_account?ref=v2.1.13"
+  count                    = var.nodo_pagamenti_test_enabled ? 1 : 0
+
+  name                     = replace(format("%s-nodotestsa", local.project), "-", "")
+  account_kind             = "BlobStorage"
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
```
pasqualespica, line 22: @aomegax we need to see how to set public $web container.
@pasqualedevita using something like that
```

resource "azurerm_storage_container" "banks" {
  depends_on = [module.nodo_test_storage[0]]

  name                  = "$web"
  storage_account_name  = module.nodo_test_storage[0].name
  container_access_type = "private"
}
```
we obtain resource already exist

pasqualespica (commented):
👍

PR #116, opened 2022-01-19
Add private dns zone for platform.pagopa.it

infantesimone:
### List of changes

- Add private dns zone for `platform.pagopa.it` in order to allow the components in `vnet` and `vnet_integration` to resolve `api.platform.pagopa.it` with the apim ip.

### Motivation and context

The goal of this PR is to allow `pagopa-proxy` (but also other components) to resolve `api.platform.pagopa.it` with apim ip.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/dns_private.tf
```diff
@@ -41,3 +41,39 @@ resource "azurerm_private_dns_zone_virtual_network_link" "privatelink_redis_cach
 
   tags = var.tags
 }
+
+resource "azurerm_private_dns_zone" "apim_private_dns_zone" {
```
pasqualedevita, line 45: Better call platform_private_dns_zone

pasqualedevita (changes requested):
Top @infantesimone only change resource name and other references

src/dns_private.tf
infantesimone: done

pasqualespica (approved):
🚀 great @infantesimone in this way all applications in to main `vnet` could call apim service directly 💪

PR #122, opened 2022-01-21
Allow developers kv access in prod and uat

uolter:
### Motivation and context

Any issue to allow developers to see and change secrets in the kv in production and uat?
Actually they can see them throughout the app service app settings.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/security.tf
+++ b/src/security.tf
@@ -70,7 +70,6 @@ data "azuread_group" "adgroup_developers" {
 
 ## ad group policy ##
 resource "azurerm_key_vault_access_policy" "adgroup_developers_policy" {
-  count = var.env_short == "d" ? 1 : 0
 
   key_vault_id = module.key_vault.id
 


src/security.tf
```diff
@@ -70,7 +70,6 @@ data "azuread_group" "adgroup_developers" {
 
 ## ad group policy ##
 resource "azurerm_key_vault_access_policy" "adgroup_developers_policy" {
-  count = var.env_short == "d" ? 1 : 0
```
pasqualespica, line 73: in this way any developer could be change prod kv too.
Are we sure it's right ? Or only sit and uat.
IMHO this give us many power but new developer too, could by mistake (ex. not paying attention to environment ) change the configuration in production.

uolter: @pasqualespica  I've assigned you and Simone the owner role to all the subscriptions.
This PR is no more needed.

PR #127, opened 2022-01-25
update ec list "verificatore UAT"

corvettafabio:
### List of changes

Add new values in APIM's "named values"

### Motivation and context

With the named value ecblacklist (nodo_pagamenti_ec) we configure the Creditor Insitutions that don't need the verify function (see activatePaymentNotice policy on APIM)

### Type of changes

- [ ] Add new resources

- [X] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (approved):
👍

PR #131, opened 2022-01-28
Update policy api-config

jacopocarlini:
### List of changes

- updated openapi for api-config
- added an exception for POST /check/xsd in the Apim policy

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (dismissed):
well 💪  
_Note_ : I know the purpose of the change, but it's always better to put it in PR description

PR #135, opened 2022-01-31
[IO-PAY-433] Add proxy multichannel support (APIM) 

MarcoAGreco:
### List of changes

- Add IO-be subscription key as secret
- Add IO-be named values into APIM
- Extend Checkout 2.0 base policy to perform binding `key <-> channel`
### Motivation and context
- Add multichannel support [RFC](https://pagopa.atlassian.net/wiki/spaces/I/pages/462193031/RFC+Implementazione+canali+multipli+su+pagopa-proxy)

### Type of changes

- [X] Add new resources

- [X] No

- [ ] Yes
- [X] No

### Other information

---


src/api/checkout/checkout_payment_activations/v1/_base_policy.xml.tpl
```diff
@@ -18,6 +18,18 @@
       <rate-limit-by-key calls="150" renewal-period="10" counter-key="@(context.Request.Headers.GetValueOrDefault("X-Forwarded-For"))" />
     </inbound>
     <outbound>
+        <!-- Handle API-key - channel binding -->
+        <when condition="@(context.Request.HeadersGetValueOrDefault("Ocp-Apim-Subscription-Key","").Equals({{io-backend-api-key}}))">
```
infantesimone, line 22: sounds good 😁 but the IO check should be placed in a new API group

pasqualespica (dismissed):
🚀

PR #139, opened 2022-02-02
[PMF-13] Add FdR function

pasqualespica:
The purpose of this PR is to add a FdR function service to read [nodoInviaFlussoRendicontazione](https://pagopa.atlassian.net/wiki/spaces/PAG/pages/211585234/nodoInviaFlussoRendicontazione), parsing it and publish on pagoPA's event hub as several `json` events

### List of changes

Added : 
  - _Flussi Di Rendicontazione_ `reporting_function.tf`

Updated : 
  - update Fdr apim policy

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/reporting_function.tf
+++ b/src/reporting_function.tf
@@ -1,0 +1,144 @@
+resource "azurerm_resource_group" "reporting_fdr_rg" {
+  count = var.reporting_fdr_enabled ? 1 : 0
+
+  name     = format("%s-reporting-fdr-rg", local.project)
+  location = var.location
+
+  tags = var.tags
+}
+
+# Subnet to host reporting-fdr function
+module "reporting_fdr_function_snet" {
+  count                                          = var.reporting_fdr_enabled && var.cidr_subnet_reporting_fdr != null ? 1 : 0
+  source                                         = "git::https://github.com/pagopa/azurerm.git//subnet?ref=v1.0.51"
+  name                                           = format("%s-reporting-fdr-snet", local.project)
+  address_prefixes                               = var.cidr_subnet_reporting_fdr
+  resource_group_name                            = azurerm_resource_group.rg_vnet.name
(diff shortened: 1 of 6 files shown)


src/reporting_function.tf
```diff
@@ -0,0 +1,139 @@
...
+
+  name     = format("%s-reporting-fdr-rg", local.project)
+  location = var.location
+
+  tags = var.tags
+}
+
+# Subnet to host reporting-fdr function
+module "reporting_fdr_function_snet" {
+  count                                          = var.reporting_fdr_enabled && var.cidr_subnet_reporting_fdr != null ? 1 : 0
```
uolter, line 12: Is it really required this flag **reporting_fdr_enabled** ?

src/reporting_function.tf
```diff
@@ -0,0 +1,139 @@
...
+    name = "default"
+    service_delegation = {
+      name    = "Microsoft.Web/serverFarms"
+      actions = ["Microsoft.Network/virtualNetworks/subnets/action"]
+    }
+  }
+}
+
+module "reporting_fdr_function" {
+  count  = var.reporting_fdr_enabled ? 1 : 0
```
uolter, line 30: As per my comment above .... do we really really need this flag **reporting_fdr_enabled**.
Why not just create the function in all envs?

uolter (changes requested):

src/reporting_function.tf
pasqualespica: no you're right i'll remove it

src/reporting_function.tf
pasqualespica: see above response : )

uolter (approved):

PR #156, opened 2022-02-08
[PM-382] metadata

infantesimone:
### List of changes

- add metadata spid for WISP (Peyment Manager)

### Motivation and context

The goal of this PR is to expose the SPID metadata for WISP - Payment Manager.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/apim_payment_manager.tf
```diff
@@ -478,6 +478,7 @@ locals {
     path                  = "wallet"
     subscription_required = false
     service_url           = null
+    metadata              = var.env_short == "p" ? file("./api/payment_manager_api/wisp/sp_metadata_PROD.xml") : file("./api/payment_manager_api/wisp/sp_metadata_UAT.xml")
```
pasqualedevita, line 481:
```suggestion
    metadata              = file("./api/payment_manager_api/wisp/sp_metadata_${var.env}.xml")
```
u can set/create a var.env for more simple sintax.

pasqualedevita (dismissed):
I saw some certificates in policy, are they safe?

src/apim_payment_manager.tf
pasqualedevita: not blocking

infantesimone:
> I saw some certificates in policy, are they safe?

It is a temporary solution, the final solution should be saved in a storage of the Azure PM. Anyway, you are right, better save it in the kv. thank you very much! @pasqualedevita

pasqualedevita (approved):

PR #184, opened 2022-03-07
Add option to set additional origins for checkout APIs

giovanniberti:
### List of changes

* Add option to set additional origins for checkout APIs

### Motivation and context
This PR is necessary to allow the checkout frontend instance deployed on the DEV environment to use UAT backend.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/apim_checkout.tf
+++ b/src/apim_checkout.tf
@@ -92,7 +92,8 @@ module "apim_checkout_payments_api_v1" {
   })
 
   xml_content = templatefile("./api/checkout/checkout_payments/v1/_base_policy.xml.tpl", {
-    origin = format("https://%s.%s/", var.dns_zone_checkout, var.external_domain)
+    origin                 = format("https://%s.%s/", var.dns_zone_checkout, var.external_domain)
+    additional_origin_tags = var.env_short == "u" ? format("<origin>https://dev.checkout.%s/</origin>", var.external_domain) : ""
   })
 }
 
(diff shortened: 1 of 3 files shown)


src/apim_checkout.tf
```diff
@@ -92,7 +92,8 @@ module "apim_checkout_payments_api_v1" {
   })
 
   xml_content = templatefile("./api/checkout/checkout_payments/v1/_base_policy.xml.tpl", {
-    origin = format("https://%s.%s/", var.dns_zone_checkout, var.external_domain)
+    origin                 = format("https://%s.%s/", var.dns_zone_checkout, var.external_domain)
+    additional_origin_tags = var.env_short == "u" ? format("<origin>https://dev.checkout.%s/</origin>", var.external_domain) : ""
```
infantesimone, line 96: would leave the `origin` tag in the policy file, what do you think?

giovanniberti: I tried something like `<origin>${additional_origin}</origin>` and Azure doesn't like it at all when the origin tag is empty (I don't like putting XML as an argument either). We could use a list as an agument and use a [for loop](https://www.terraform.io/language/functions/templatefile#lists) inside the template, but it could be a bit overkill for this, what do you think?

uolter (approved):

src/apim_checkout.tf
infantesimone: @giovanniberti we can close this PR, we will no longer have `dev-checkout` <--> `uat-PM`

PR #191, opened 2022-03-09
[PPD-105] Add vars to GPD

aomegax:
### List of changes
Added vars to GPD Reporting function:

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/variables.tf
+++ b/src/variables.tf
@@ -1091,6 +1091,24 @@ variable "gpd_cron_schedule_expired_to" {
   default     = null
 }
 
+variable "gpd_max_retry_queuing" {
+  type        = string
+  description = "Max retry queuing when the node calling fails."
+  default     = 5
+}
+
+variable "gpd_queue_retention_sec" {
+  type        = string
+  description = "The maximum time to allow the message to be in the queue."
+  default     = 86400
+}
+
+variable "gpd_queue_delay_sec" {
+  type        = string
+  description = "The length of time during which the message will be invisible, starting when it is added to the queue."
+  default     = 3600
+}
+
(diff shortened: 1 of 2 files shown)


src/variables.tf
```diff
@@ -1091,6 +1091,24 @@ variable "gpd_cron_schedule_expired_to" {
...
 }
 
+variable "gpd_max_retry_queuing" {
+  type        = string
+  description = "Max retry queuing when the node calling fails."
+  default     = 5
+}
+
+variable "gpd_queue_retention_sec" {
+  type        = string
```
uolter, line 1101:
```suggestion
  type        = number
```

src/variables.tf
```diff
@@ -1091,6 +1091,24 @@ variable "gpd_cron_schedule_expired_to" {
   default     = null
 }
 
+variable "gpd_max_retry_queuing" {
+  type        = string
```
uolter, line 1095:
```suggestion
  type        = number
```

src/variables.tf
```diff
@@ -1091,6 +1091,24 @@ variable "gpd_cron_schedule_expired_to" {
...
+}
+
+variable "gpd_queue_retention_sec" {
+  type        = string
+  description = "The maximum time to allow the message to be in the queue."
+  default     = 86400
+}
+
+variable "gpd_queue_delay_sec" {
+  type        = string
```
uolter, line 1107:
```suggestion
  type        = number
```

uolter (changes requested):

PR #197, opened 2022-03-14
Fix for GPD deployment

jacopocarlini:
The purpose of this PR is to fix configuration parameters for GPD 

### List of changes

Upd : 
  - GPD host reference
  - `SPRING_DATASOURCE` configuration

Add :  
  - `flyway_gpd.sh` script to create `apd` schema and `apd` user

**NOTE** : this PR foresees that on KV they are present :  
   - `db-apd-user-name`
   -  `db-apd-user-password`
   -  `gpd-paa-password`

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


uolter (dismissed):

uolter (dismissed):

src/main.tf
```diff
@@ -8,7 +8,10 @@ terraform {
       source  = "hashicorp/azuread"
       version = "= 2.6.0"
     }
-
+    postgresql = {
+      source  = "cyrilgdn/postgresql"
```
uolter, line 12: is it safe to use this provider? Is it kind of ufficial?

src/postgresql.tf
```diff
@@ -9,13 +9,13 @@ module "postgresql_snet" {
 }
 
 data "azurerm_key_vault_secret" "db_administrator_login" {
-  count        = var.prostgresql_enabled ? 1 : 0
+  count        = var.postgresql_enabled ? 1 : 0
```
uolter, line 12: Do we really need this count .... why not just create the resource in every environment?

pasqualespica: unfortunately in this case yes.
In the dev environment we'll have a single server instance (used by other applications such as the mock psp too cc @infantesimone ) while in uat / prod we'll go towards a [Flexible solution](https://pagopa.atlassian.net/wiki/spaces/PAG/pages/474219118/Migrazione+DB+Oracle+su+Cloud+Azure#Tipologia-di-DB-PaaS) cc @pasqualedevita 

I changed the condition removed `postgresql_enabled`  feature flag and user `env_short` clearly
do u agree @uolter  ?

src/main.tf
pasqualespica: I removed this provider and we decided to create user/grant postegresql via flyway script

uolter (approved):

PR #216, opened 2022-03-28
Add PPT-LMI  SYNC FATTURAZIONE  WEB-BO WFESP

pasqualespica:
The purpose of this PR is to add on apim all nodo oncloud services

### List of changes

Add on apim nodo's services : 
- ppt-lmi
- sync
- web-bo
- wfesp

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/variables.tf
```diff
@@ -225,12 +225,25 @@ variable "ip_nodo" {
   description = "Nodo pagamenti ip"
 }
 
-variable "nodo_pagamenti_oncloud_url" {
+variable "lb_aks" {
   type        = string
-  description = "Nodo pagamenti url oncloud"
+  description = "Load balancer AKS Nexi/SIA"
   default     = "http://"
```
uolter, line 231: mmmm this default it's a bit confusing don't you?

uolter (approved):

src/variables.tf
pasqualespica: ok,  I fix it

PR #260, opened 2022-05-18
[PPR-341] add cosmosdb account and first db

pasqualespica:
The purpose of this PR is to add cosmos db account and a first db according to [nodo-dei-pagamenti-biz-evt](https://pagopa.atlassian.net/wiki/spaces/PAG/pages/497746877/Design+Review-+Evento+di+pagamento) 
and new logicapp
image

### List of changes

Add : 
  - cosmos db account
  - document db `paymentsdb`
  - new subnet
  - private endpoint

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/dns_private.tf
```diff
@@ -102,3 +102,21 @@ resource "azurerm_private_dns_zone" "privatelink_azurecr_pagopa" {
 
   tags = var.tags
 }
+
+# MongoDB
+resource "azurerm_private_dns_zone" "privatelink_mongo_cosmos_azure_com" {
```
infantesimone, line 107: private link for all envs? also for DEV?

src/mongodb.tf
```diff
@@ -0,0 +1,141 @@
...
+
+  allowed_virtual_network_subnet_ids = []
+
+  tags = var.tags
+}
+
+# database(s)
+
+# 1) payments db
+resource "azurerm_cosmosdb_mongo_database" "pagopa_payments_cosmosdb_mongodb_db" {
```
infantesimone, line 66: could it be useful to decouple the cosmos  account definition from the product-specific  database definition?

src/dns_private.tf
pasqualespica: disable for DEV

src/mongodb.tf
pasqualespica: do you say to put the db and its collection in the separate TF file?

infantesimone: yes, but it's just an idea. We can also refactor in the future (if the number of databases/collections increase).

pasqualespica: good tip, i put there into dedicated file

PR #266, opened 2022-05-20
[PPD-175] add reporting autoscaling

pasqualespica:
The purpose of this PR is to fix [reporting](https://pagopa.atlassian.net/wiki/spaces/PPD/pages/456525211/Design+Review+-+Rendicontazione) az-fn autoscaling.

### List of changes

Add `azurerm_monitor_autoscale_setting` for az function : 
  - `reporting_batch_function`
  - `reporting_service_function`
  - `reporting_analysis_function`

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


aomegax:
Due to shared plan, it can happen one function should scale up and another one should scale down. This should bring instability to the system. 
I'd like to find a way to avoid this behavior. What do you think about?

PR #336, opened 2022-07-29
feat(apim): add apim product and api for `pagopa-noptifications-service`

giovanniberti:
### List of changes

* add apim product and api for `pagopa-noptifications-service`

### Motivation and context

This PR adds APIs to Azure APIM to support the new `pagopa-notifications-service`

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/shared/04_apim_pagopa_notifications_service.tf
```diff
@@ -0,0 +1,66 @@
...
+}
+
+##############################
+## API notifications service ##
+##############################
+locals {
+  apim_pagopa_notifications_service_api = {
+    display_name          = "pagoPA Shared - notifications service API"
+    description           = "API to support notifications service"
+    path                  = "shared/notifications-service"
```
infantesimone, line 26: For the basepath I would leave only the product name, without the domain reference, so: 

- shared/notifications-service --> notifications-service

what do you think?

src/shared/04_apim_pagopa_notifications_service.tf
```diff
@@ -0,0 +1,66 @@
...
+    subscription_required = false
```
infantesimone, line 27: The api key is required

infantesimone (approved):

PR #362, opened 2022-08-30
fix: Split deny path strings

pasqualespica:
The purpose of this pr is to overcome size limit of app gtw rewrite rules

> needs to be less than 512 characters. For more details refer to the documentation here: https://aka.ms/appgwheadercrud."

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


infantesimone (dismissed):
great! maybe `app_gateway_deny_paths_2` rather than `app_gateway_deny_paths2`?

PR #429, opened 2022-09-29
feat: Add Grafana Managed

rtrive:
### List of changes
Add grafana managed and cname record

### Motivation and context
Gilda k6 resources

### Type of changes

- [x] Add new resources

- [x] No

- [x] Yes
- [x] No

### Other information

---


src/core/grafana.tf
```diff
@@ -0,0 +1,18 @@
+
+
+resource "azurerm_dashboard_grafana" "example" {
+  name                              = format("%s-grafana", local.project)
```
diegolagospagopa, line 4: Please use the interpolation format

src/domains/load-test/01_grafana.tf
```diff
@@ -0,0 +1,21 @@
+module "grafana_managed" {
+  source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//grafana?ref=feature/grafana"
+
+  name = format("%s-grafana", var.env_short)
```
diegolagospagopa, line 4: please use interpolation

src/domains/load-test/01_grafana.tf
```diff
@@ -0,0 +1,21 @@
...
+
+  resource_group_name = var.terraform_remote_state_core.resource_group_name
```
diegolagospagopa, line 6: Please put the grafana in its own resource group, and not use the one used by terraform state

diegolagospagopa (changes requested):

src/domains/load-test/01_grafana.tf
rtrive: I created a new one. Is it right?

src/core/grafana.tf
rtrive: I copied from all the code that is inside the repo; example: https://github.com/pagopa/pagopa-infra/blob/4243840adcf9bc46444cfe27588c6e6d3900000e/src/core/apim.tf#L4.

Is interpolation still ok?

pasqualespica (approved):
great work 🚀 
I'd only mention in PR description use of `terraform-azurerm-v3.git`

src/domains/load-test/01_grafana.tf
```diff
@@ -0,0 +1,28 @@
...
+
+  resource_group_name = azurerm_resource_group.load_test.name
+
+  tags = var.tags
+
+}
+
+resource "azurerm_dns_cname_record" "grafana" {
+
+  name                = "grafana.${var.dns_zone_prefix}.${var.external_domain}"
```
diegolagospagopa, line 21: This is only the record dns, not the entire fqdn

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/dns_cname_record

diegolagospagopa:
```suggestion
name                = "grafana"
```

diegolagospagopa:
Good job with the refactors, i have one question. This grafana will be used only for load-test?

if not. maybe i think is better to call the resource group/folders/state with another name.

src/domains/load-test/01_grafana.tf
diegolagospagopa: yes

src/core/grafana.tf
diegolagospagopa: yes it is

rtrive:
> Good job with the refactors, i have one question. This grafana will be used only for load-test?
> 
> if not. maybe i think is better to call the resource group/folders/state with another name.

Got your point, I used `load-test` because the scope is limited to the `gilda` and I don't know if Grafana will be useful for others

diegolagospagopa:
> > Good job with the refactors, i have one question. This grafana will be used only for load-test?
> > if not. maybe i think is better to call the resource group/folders/state with another name.
> 
> Got your point, I used `load-test` because the scope is limited to the `gilda` and I don't know if Grafana will be useful for others

ok perfect

diegolagospagopa (approved):

diegolagospagopa:
Great work, i think is ok and can be merged

PR #440, opened 2022-10-03
feat: Add custom domain prf cert on apim

infantesimone:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/core/apim.tf
+++ b/src/core/apim.tf
@@ -22,6 +22,7 @@ locals {
   portal_cert_name_proxy_endpoint = format("%s-proxy-endpoint-cert", "portal")
 
   api_domain        = format("api.%s.%s", var.dns_zone_prefix, var.external_domain)
+  prf_domain        = format("prf.%s.%s", var.dns_zone_prefix, var.external_domain)
   portal_domain     = format("portal.%s.%s", var.dns_zone_prefix, var.external_domain)
   management_domain = format("management.%s.%s", var.dns_zone_prefix, var.external_domain)
 }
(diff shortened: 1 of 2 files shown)


pasqualedevita (changes requested):
`prf` is only for `UAT` env... So I think you need to use `count` and `dynamic` structures to prevent errors in `DEV` and `PROD` envs

src/core/apim.tf
```diff
@@ -572,6 +573,15 @@ resource "azurerm_api_management_custom_domain" "api_custom_domain" {
     )
   }
 
+  proxy {
+    host_name = local.prf_domain
+    key_vault_id = replace(
+      data.azurerm_key_vault_certificate.app_gw_platform.secret_id,
+      "/${data.azurerm_key_vault_certificate.app_gw_platform.version}",
```
pasqualedevita, line 580: this certificate contains the common name `api.<env>.platform.pagopa.it`

pasqualedevita (changes requested):

src/core/apim.tf
infantesimone: > `prf` is only for `UAT` env... So I think you need to use `count` and `dynamic` structures to prevent errors in `DEV` and `PROD` envs

I created a separate `azurerm_api_management_custom_domain` only in uat env , okay?

infantesimone: @pasqualedevita is it okay to use dynamic to have it only in uat?

PR #466, opened 2022-10-16
feat: api logs without body

infantesimone:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/core/api_logs.tf
+++ b/src/core/api_logs.tf
@@ -18,6 +18,20 @@ locals {
     module.apim_api_gpd_api.name,
   ]
 
+  api_info_log = [
+    module.apim_pm_restapi_api_v4.name,
+    module.apim_pm_restapirtd_api_v1.name,
+    module.apim_pm_restapirtd_api_v2.name,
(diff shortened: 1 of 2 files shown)


pasqualespica (dismissed):
💪

src/core/api_logs.tf
```diff
@@ -52,3 +66,36 @@ resource "azurerm_api_management_api_diagnostic" "apim_logs" {
...
+    body_bytes = 0
+  }
+
+  backend_request {
+    body_bytes = 0
+  }
+
+  backend_response {
+    body_bytes = 0
+  }
```
pasqualedevita, line 100:
```suggestion
```
according to https://registry.terraform.io/providers/hashicorp/azurerm/2.99.0/docs/resources/api_management_api_diagnostic#backend_request

frontend_request, frontend_response, backend_request, backend_response are optional so maybe better remove them.

Anyway setting to 0 body request are not saved into logs.

Since PM is critical application a test in DEV/UAT can be useful

PR #474, opened 2022-10-21
[PGEC-83] Add touchpoint container

MarcoAGreco:
### List of changes

- Add cosmosDB container for `afm-marketplace-be`

### Motivation and context

- Support touchpoint persistency

### Type of changes

- [ ] Add new resources

- [X] No

- [ ] Yes
- [X] No

### Other information

---

### DIFF
--- a/src/domains/afm-common/03_cosmosdb_marketplace.tf
+++ b/src/domains/afm-common/03_cosmosdb_marketplace.tf
@@ -107,7 +107,7 @@ locals {
       partition_key_path = "/idPsp",
     },
     {
-      name = "touchpoints",
+      name               = "touchpoints",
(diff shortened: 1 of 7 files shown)


src/domains/afm-common/03_cosmosdb_marketplace.tf
```diff
@@ -106,6 +106,10 @@ locals {
       name               = "validbundles",
       partition_key_path = "/idPsp",
     },
+    {
+      name               = "touchpoints",
+      partition_key_path = null
```
aomegax, line 111: add partition key, it is required

aomegax (changes requested):

aomegax (approved):

PR #494, opened 2022-11-02
Fix: Grafana CNAME

rtrive:
### List of changes
Fix wrong cname

### Motivation and context
Wrong DNS prefix and output from grafana managed module

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

Depends on: https://github.com/pagopa/terraform-azurerm-v3/pull/4

---

### DIFF
--- a/src/domains/load-test/01_grafana.tf
+++ b/src/domains/load-test/01_grafana.tf
@@ -6,11 +6,13 @@ resource "azurerm_resource_group" "load_test" {
 }
 
 module "grafana_managed" {
-  source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//grafana?ref=v3.1.0"
+  source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//grafana?ref=feature/new_output_grafana"
 
   name = "${local.product}-grafana"
 
   resource_group_name = azurerm_resource_group.load_test.name
+
+  api_key_enabled = true
 
   tags = var.tags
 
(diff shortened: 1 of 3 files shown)


src/domains/load-test/env/weu-dev/terraform.tfvars
```diff
@@ -23,5 +23,5 @@ terraform_remote_state_core = {
   key                  = "dev.terraform.tfstate"
 }
 
-dns_zone_prefix = "dev.platform"
+dns_zone_prefix = "dev.platform.pagopa.it"
```
pasqualespica, line 26: why in `dns_zone_prefix` y'd add `.pagopa.it` too ?

src/domains/load-test/01_grafana.tf
```diff
@@ -22,7 +22,7 @@ resource "azurerm_dns_cname_record" "grafana" {
   zone_name           = var.dns_zone_prefix
   resource_group_name = local.vnet_resource_group_name
   ttl                 = 3600
-  record              = module.grafana_managed.endpoint
+  record              = module.grafana_managed.hostname
```
pasqualespica, line 25: if it's related to https://github.com/pagopa/terraform-azurerm-v3/pull/4 mention it with `depends_on` into PR description

src/domains/load-test/env/weu-dev/terraform.tfvars
rtrive: because without it, it gives me an error

PR #518, opened 2022-11-11
[BizEvents] feat: PPR-497 add cosmos db key secret in kv

AngeloCaporaso:
### List of changes

- Add the cosmos db key secret in key vault definition
- Add apim for biz events service, policy and openapi

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
great work @AngeloCaporaso 💪

PR #566, opened 2022-12-01
[PAGOPA-67] New Container CDI

jacopocarlini:
### List of changes

added new container

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


aomegax (dismissed):
👍

src/core/security.tf
```diff
@@ -238,6 +238,15 @@ data "azurerm_key_vault_secret" "db_nodo_pwd" {
   key_vault_id = module.key_vault.id
 }
 
+data "azurerm_key_vault_secret" "cosmos_key" {
```
aomegax, line 241: add the related project, because in `core` could be confusing. 

besides I prefer that the keyvault is created by infra however even if the value is set manually (i.e. [here](https://github.com/pagopa/pagopa-infra/blob/PAGOPA-67-gec-storico-cdi-da-caricare-su-afm/src/domains/afm-common/02_security.tf#L86))

aomegax (changes requested):
small fixes

src/core/security.tf
jacopocarlini: done 👍

aomegax (approved):
👍

aomegax (approved):
👍

PR #606, opened 2023-01-03
refactor(checkout): move checkout APIs for ECs to separate APIM product

giovanniberti:
### List of changes

* Move checkout APIs for ECs to separate APIM product

### Motivation and context

Right now all Checkout APIs are currently under the same APIM product. We want to publish the new Checkout Carts API, but publishing options for APIs can only be applied product-wise, and currently the Checkout product contains APIs that have no reason to be published (e.g. `CdInfoWisp`). 
This PR separates the Checkout Cart API into a separate product meant for EC consumption.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/core/apim_checkout_ec.tf
```diff
@@ -0,0 +1,68 @@
+##############
+## Products ##
+##############
+
+module "apim_checkout_ec_product" {
+  count  = var.checkout_enabled ? 1 : 0
```
infantesimone, line 6: Is this count needed?

infantesimone: ping :-)

giovanniberti: I left it in for consistency with current checkout configuration in [`apim_checkout.tf`](https://github.com/pagopa/pagopa-infra/blob/cb5eb907f66c808db4fde0f58e12d502b4263615/src/core/apim_checkout.tf)

infantesimone: if we know that this APIs will be on all environments, better remove this count

infantesimone (approved):

PR #620, opened 2023-01-10
PAGOPA-578 feat: adding poc apim and sa

FedericoRuzzier:
Co-authored-by: Angelo Caporaso <AngeloCaporaso@users.noreply.github.com>

### List of changes
Addition of apim and storage account for the poc for both micronaut and quarkus. 

### Motivation and context
To proceed in the development of the application. 

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/shared-app/04_apim_reporting_orgs_enrollment.tf
```diff
@@ -0,0 +1,75 @@
+###########################################
+## Product POC REPORTING ORGS ENROLLMENT ##
+###########################################
+
+locals {
+  apim_poc_service_api = {
+    display_name          = "Reporting orgs enrollment - POC pagoPA"
+    description           = "API to support POC"
+    path                  = "shared/poc-reporting-orgs-enrollment"
```
aomegax, line 9: I suggest to use this path `shared/poc/reporting-orgs-enrollemnt` in order to reuse `shared/poc` for future needs

src/domains/shared-app/04_apim_reporting_orgs_enrollment.tf
```diff
@@ -0,0 +1,75 @@
...
+    subscription_required = true
+    service_url           = null
+  }
+}
+
+##############
+## Products ##
+##############
+
+module "apim_poc_product" {
```
aomegax, line 19: I think we should have poc only in dev...so we should use `count` to apply them only in that env

FedericoRuzzier:
Do not merge pr yet, still in progress

pasqualespica (approved):

PR #686, opened 2023-01-30
feat: Add afm Storage Account PAGOPA-641

AngeloCaporaso:
### List of changes

- AFM storage account addition
- Issuer range table addition
- Storage connection string added in the Key Vault

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


jacopocarlini (approved):
LGTM

PR #760, opened 2023-02-24
Feature/paselc 74

biaMurNtt:
### List of changes

Update open api for APIM api

### Motivation and context

update api parameter

### Type of changes

- [ ] Add new resources

- [X] No

- [ ] Yes
- [X] No

### Other information

---


pasqualespica:
in agreement to @biaMurNtt  we close it because it isn't needed, replaced by https://github.com/pagopa/pagopa-infra/pull/964

PR #770, opened 2023-03-03
Chk 1193 ecommerce queues refactoring

pietro-tota:
### List of changes
Refactoring of all ecommerce queues to change from an `queue for event` approach to `queue for behaviour`approach.

### Motivation and context
Those modifications are needed in order to clarify the scope of each queue

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


infantesimone (changes requested):
let's try with a backward compatible PR, leave the old queues and then delete them when we switch all the ecommerce services (in another PR)

pietro-tota:
> let's try with a backward compatible PR, leave the old queues and then delete them when we switch all the ecommerce services (in another PR)

Fixed, I've added also the queue present in this PR pagopa/pagopa-infra#759

infantesimone (approved):

PR #848, opened 2023-03-31
[PAGOPA-841] fix: set reporting analysis subkey validation

andrea-deri:
During the integration tests performed on GPD-Reporting applications, it was found that the APIs for GPD-Reporting-Analysis services were not protected by a subscription key: they are invocable directly without using any particular key. In particular, for the `getFlow` API, a missing configuration in code mandated the use of a function-key on the Azure Function that allowed any unwanted access to be blocked.

### List of changes
 - Added subscription key securization on `getFlow` and `getFlowList` endpoints

### Motivation and context
This changes are required in order to protect the described APIs

### Type of changes
- [ ] Add new resources

- [ ] No

- [ ] Yes
- [X] No

### Other information

---

### DIFF
--- a/src/core/.terraform.lock.hcl
+++ b/src/core/.terraform.lock.hcl
@@ -5,6 +5,7 @@ provider "registry.terraform.io/azure/azapi" {
   version     = "1.3.0"
   constraints = "1.3.0"
   hashes = [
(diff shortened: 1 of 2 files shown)


jacopocarlini (dismissed):
👍

pasqualespica (approved):

PR #854, opened 2023-04-03
[CHK-1348]add optional fields to cart request

ciuffagianluca:
Add `idCart` and `allCCP` fields to cart request

### List of changes

Add these fields to the body of the request. Both are optional

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/core/api/checkout/checkout_ecommerce/v1/_openapi.json.tpl
```diff
@@ -1448,6 +1448,14 @@
                 "example": "www.comune.di.prova.it/pagopa/error.html"
               }
             }
+          },
```
pietro-tota, line 1451: Are we still updating this api group? As I know this one is marked for removal waiting for frontend upgrade in UAT, so I will not upgrade also this API group upgrading only the one into ecommerce-app domain.
What do you think?

pietro-tota (approved):
Just a comment

src/core/api/checkout/checkout_ecommerce/v1/_openapi.json.tpl
ciuffagianluca: You're right. I updated to be 100% compliant

ciuffagianluca: Removed

pietro-tota (approved):

src/core/api/checkout/checkout_ec/v1/_openapi.json.tpl
```diff
@@ -159,6 +159,14 @@
                 "example": "www.comune.di.prova.it/pagopa/error.html"
               }
             }
+          },
```
infantesimone, line 162: to delete

infantesimone (changes requested):

infantesimone (approved):

PR #860, opened 2023-04-05
feat: [PAGOPA-796] Authorizer - Multi-entity Sub-Key caching authorizator

andrea-deri:
The Authorizer will be a new system that permits to a client to be authorized to execute one or more operations on certain entity (creditor institution, PSP, etc) in a simplified and centralized way using a caching system directly by the APIM component. The client that has a valid subscription key can be authorized to performs operation on a set of entities for one or more domains (i.e. GPD, APIConfig, AFM, etc) and will be forbidden to perform operations on the other entities on which it does not has visibility or other domains on which it cannot have the related subscription key.
This PR contains all the new modules and resources in order to build the infrastructure for the Authorizer system.

### List of changes
 - Added a new function app that will contains several functions that will save and update the authorizations orchestrating the communication with APIM
 - Added a CosmosDB used for saving the relation Domain-subkey-entities that will trigger the functions in order to perform the authorizations sync process
 - Added a set of internal APIs that will be called by Cache's functions that will be used to execute the authorizations sync process

### Motivation and context
This changes are required in order to define the main structure for the Authorizer system

### Type of changes
- [X] Add new resources

- [X] No

- [ ] Yes
- [X] No

### Other information

---


pasqualespica (approved):

pasqualespica (approved):

pasqualespica (dismissed):
great

pasqualespica (commented):
great work 👍

PR #862, opened 2023-04-05
fix: xpay psp list

pietro-tota:
### List of changes
Add PSP with id `60000000001` into XPAY psp list

### Motivation and context
This modification is needed in order to configure psp for XPAY correctly

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


ciuffagianluca (approved):
Well done!

PR #867, opened 2023-04-11
CHK-1363-auth-request

Fernando-Granato:
### List of changes
Changed http response code from 400 to 504 into auth-request base policy.

### Motivation and context
Changed the policy for pspId check. In case the pspId was not registered on the XPAY or VPOS lists it will return a 504

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
```diff
@@ -45,14 +45,14 @@
             </when>
             <otherwise>
                 <return-response>
-                    <set-status code="400"/>
+                    <set-status code="504"/>
```
infantesimone, line 48: why is it a timeout in this case?

infantesimone (changes requested):

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
infantesimone: I think the 4XX is correct as the input is not valid

Fernando-Granato: OK. So the task can be closed ?

Fernando-Granato: Or we can discuss it at a refinement

pietro-tota: Maybe a 422 is more suitable for this specific case

Fernando-Granato: can be a good idea! I report your idea on JIRA 🙂

PR #909, opened 2023-05-01
feat: Added Cert mounter middleware 

diegolagospagopa:
### List of changes

- added cert-mounter
- added status-app
- added template yaml for both helms

### Motivation and context

Allow to have always cert-mounter in yuor installation

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


infantesimone:
@diegolagospagopa  cert-mounter added in this PR, is it equivalent to the one like https://github.com/pagopa/pagopa-infra/blob/83c80c2e207273b26f69e362b363a3710c5f5a36/src/domains/wallet-app/02_namespace.tf#LL36C2-L36C2 ?
if there are both we can have conflicts
cc @pietro-tota

PR #923, opened 2023-05-04
[PAGOPA-925] fix: resolved bug on APIM-Redis connectivity

andrea-deri:
During the testing phase in UAT environment of the `pagopa-platform-authorizer` service, it was found that APIM is unable to communicate properly with an external Redis cache. This is caused by a misconfiguration of the Redis' private DNS zone that did not allow connection between APIM (present in `vnet-integration` VNET) and Redis itself (present in `vnet` core VNET).
In order to resolve this issue, it was made an update either on APIM and on external cache, adding two new resources:
 - one for permitting the connection between Redis' private DNS zone and the `vnet-integration` Virtual network 
 - one for permitting the generation of an explicit link between APIM and Redis, due to the fact that the default one does not work correctly on APIM uses.

In future, when a well-known [bug on APIM](https://learn.microsoft.com/en-us/answers/questions/1190454/how-to-read-from-apim-policy-an-external-redis-cac?page=1&orderby=Helpful&comment=answer-1188540#newest-answer-comment) will be resolved, the position of the Redis resource can be moved from `vnet-integration` VNET to `vnet` core VNET in order to move the responsibility to the last one. 

### List of changes
 - Added `azurerm_private_dns_zone_virtual_network_link` resource for generating a new link between vnet-integration network and private DNS zone.
 - Added `azurerm_api_management_redis_cache` resource for generating a new link between APIM and Redis external cache.

### Motivation and context
This fix is needed in order to resolve the connectivity issue between APIM and Redis cache.

### Type of changes
- [X] Add new resources

- [X] Yes, users may be impacted applying this change
- [ ] No

- [ ] Yes
- [X] No

### Other information
The apply on environment was made using the following commands:
```
sh terraform.sh apply <ENV> \
-target=azurerm_private_dns_zone_virtual_network_link.vnet_integration_network_link \
-target=azurerm_api_management_redis_cache.apim_external_cache_redis \
-target=module.apim_api_debt_positions_api_v1 \
-target=module.redis
```

---

### DIFF
--- a/src/core/README.md
+++ b/src/core/README.md
@@ -430,6 +430,7 @@
 | [azurerm_api_management_product_api.apim_nodo_dei_pagamenti_product_api](https://registry.terraform.io/providers/hashicorp/azurerm/2.99.0/docs/resources/api_management_product_api) | resource |
 | [azurerm_api_management_product_api.apim_nodo_dei_pagamenti_product_api_auth](https://registry.terraform.io/providers/hashicorp/azurerm/2.99.0/docs/resources/api_management_product_api) | resource |
 | [azurerm_api_management_product_api.apim_nodo_dei_pagamenti_product_api_dev](https://registry.terraform.io/providers/hashicorp/azurerm/2.99.0/docs/resources/api_management_product_api) | resource |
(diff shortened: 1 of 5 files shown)


pasqualespica (approved):
great work 🔝

pasqualespica:
@andrea-deri we must then understand better the problem of the redis connection string in the apim because it doesn't go through the module

PR #925, opened 2023-05-05
feat: Dis 21 migrazione dominio nodo v3

re-sh-cloud-arch:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/nodo-secret/storageaccount.tf
```diff
@@ -1,19 +1,19 @@
...
-  versioning_name            = "versioning"
-  enable_versioning          = false
-  resource_group_name        = azurerm_resource_group.sec_rg.name
-  location                   = var.location
-  advanced_threat_protection = false
-  allow_blob_public_access   = false
+  name                     = replace("${local.product}-${var.domain}-nodocerts-sa", "-", "") # nodocerts<dev|uat|prod>
+  account_kind             = "StorageV2"
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
```
diegolagospagopa, line 7: is correct that must be "LRS" and not GRS o something more?

src/domains/nodo-secret/99_locals.tf
```diff
@@ -4,4 +4,6 @@ locals {
 
   monitor_action_group_slack_name = "SlackPagoPA"
   monitor_action_group_email_name = "PagoPA"
+
+  pipe_service_principel_application_id = "b7375314-3eff-4760-9952-84f96ee669cf"
```
diegolagospagopa, line 8: can you put a comment with the name of the SP, to simplify the search pleasee

diegolagospagopa (changes requested):

src/domains/nodo-secret/storageaccount.tf
re-sh-cloud-arch: at the moment is correct. with DR implementation will be migrate to GZRS

src/domains/nodo-secret/99_locals.tf
re-sh-cloud-arch: with application ID in Azure Portal search is more quickly then with name. use name of SP for search give more than one result

fmlisco (approved):

uolter (approved):

PR #964, opened 2023-05-15
Update apim

GiacomoBrancazi:
### List of changes

update api-doc

### Motivation and context

exposed api on apim

### Type of changes

- [ ] Add new resources

- [X] No

- [ ] Yes
- [X] No

### Other information

---


jacopocarlini (changes requested):
Use a different path of this file

pasqualespica (approved):

PR #988, opened 2023-05-25
CHK-1556-added-jwt-policy-fees

Fernando-Granato:
### List of changes

- Added transactionId header reading for JWT validation.
- Added transactionId header and security token for ecommerce-checkout openapi POST fees
- Added transactionId header for ecommerce-payment-methods openapi POST fees

### Motivation and context
The change was made to validate the jwt token for POST fees.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-checkout/v1/_validate_transactions_jwt_token.tpl
```diff
@@ -3,8 +3,15 @@
         <base />
         <set-header name="x-transaction-id" exists-action="delete" />
         <set-variable name="requestTransactionId" value="@{
-            return context.Request.MatchedParameters["transactionId"];
-        }" />
+                    return context.Request.Headers.GetValueOrDefault("transactionId","");
```
pietro-tota, line 6: 🚲 Instead of having a choose maybe the transactionId retrieval code can be all be written directly into this set-variable tag, something like:
var transactionId = context.Request.MatchedParameters.GetValueOrDefault("transactionId","");
if(transactionId == ""){
transactionId = context.Request.Headers.GetValueOrDefault("transactionId","");
}
return transactionId;

and remove the below choose tag altogether

pietro-tota (dismissed):
LGMT🚀

pietro-tota (approved):
LGTM

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_validate_transactions_jwt_token.tpl
```diff
@@ -1,10 +1,14 @@
...
         <set-variable name="requestTransactionId" value="@{
-            return context.Request.MatchedParameters["transactionId"];
-        }" />
+                    var transactionId = context.Request.MatchedParameters.GetValueOrDefault("transactionId","");
+                    if(transactionId == ""){
+                        transactionId = context.Request.Headers.GetValueOrDefault("x-transaction-id","");
+                    }
+                    return transactionId;
+            }" />
+        <set-header name="x-transaction-id" exists-action="delete" />
```
Fernando-Granato, line 11: @infantesimone  the header delete should be unnecessary since then an override is done after the JWT validation.
`<set-header name="x-transaction-id" exists-action="override">`
Do you confirm ?

Fernando-Granato: ping!  🙂 @infantesimone

MarcoAGreco (approved):

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_validate_transactions_jwt_token.tpl
infantesimone: it is a check that makes the policy more secure, imagine a bug in the policy section which for some conditions does not value the header. At this point the header would not arrive at the service. Without this check, an attacker could inject a value of the a header from outside

infantesimone (approved):

PR #997, opened 2023-05-31
feat: DIS-55 Added dns forwarder backup

mamari90:
### List of changes

- removed pair regions and object because with integration vnet is impossible to use
- created packer image for dns forwarder
- created `next-core` to create resources in core with azurerm v3
- created scale set for dns forwarded vm
- fix aks platform problems
- minor fixs

added new dns forwarder related resources in pair region for DR

### Motivation and context

Allow to have a backup DNS forwarder that permit debug is the main dns forwarder is broken

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/next-core/terraform.sh
mamari90, line 1: can you link the terraform.sh stored in 'scripts' folder, or does it need to be different?

PR #1009, opened 2023-06-05
feat: Upgrade pipeline template

mamari90:
### List of changes

- upgraded pipeline template for selfcare code review
- upgraded devops agents
- created different agent scale set for infra and app pipelines (app pipelines will continue to work with the old agent pool name)

### Motivation and context

old agent was failing because it could not find some commands
old pipeline was failing because it was missing kubelogin

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

old scale set needs to be removed manually

---

applied in DEV. needs to be applied in UAT and PROD

` ./terraform.sh apply dev -target=module.azdoa_li_app -target=module.azdoa_li_infra -target=module.azdoa_li`


diegolagospagopa (approved):
Great job

PR #1031, opened 2023-06-12
[PAGOPA-997] Change on Authorizer HTTP response status

andrea-deri:
This PR contains a fix made on Authorizer fragment policy in order to resolve the wrong assignation on response HTTP status code: with this update, the output status will pass from 401 Unauthorized to 403 Forbidden, more correct on the context of the Authorizer authorization operation.

### List of changes
 - Changed HTTP response status code from 401 Unauthorized to 403 Forbidden

### Motivation and context
This change is needed in order to correctly handle the client forbidding during authorization check.

### Type of changes
- [ ] Add new resources

- [X] Yes, users may be impacted applying this change
- [ ] No

- [ ] Yes
- [X] No

### Other information

---


pasqualespica (approved):
well 👍

pasqualespica (approved):

PR #1041, opened 2023-06-14
[PAGOPA-1013] added due_date column to iban table

alessio-acitelli:
### List of changes

- added due_date column to iban table (liquibase script)

### Motivation and context

- new business request

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


jacopocarlini (changes requested):
alter table

pasqualespica (approved):

PR #1075, opened 2023-06-21
[CHK-1648] feat(storage) 1/2: add different storage account for transient and dead letter queue

ciuffagianluca:
Create storage different for different type of queues to help monitoring services

### List of changes

Added `ecommerce-storage-transient` and `ecommerce-storage-deadletter` storage account.
The second hosts all the dead letter queues, while the former hosts all the working queues.

### Motivation and context

This partition has been implemented to help monitoring queues, since the dead letter ones could be filled with no granted prcessing time, while all the working queues have predefined time to process events

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-common/03_storage.tf
```diff
@@ -43,6 +43,29 @@ resource "azurerm_private_endpoint" "storage_private_endpoint" {
...
+  name                = "${local.project}-tr-storage-private-endpoint"
+  location            = var.location
+  resource_group_name = azurerm_resource_group.storage_ecommerce_rg.name
+  subnet_id           = module.ecommerce_storage_snet.id
+  private_dns_zone_group {
+    name                 = "${local.project}-storage-private-dns-zone-group"
+    private_dns_zone_ids = [data.azurerm_private_dns_zone.storage.id]
+  }
+
+  private_service_connection {
```
infantesimone, line 59: pay attention on this configuration, i think it's ok but maybe better to investigate if we can use the same private link defined for old queue
https://learn.microsoft.com/en-us/azure/private-link/private-endpoint-overview

infantesimone (dismissed):

infantesimone (approved):

PR #1157, opened 2023-07-18
[PAGOPA-1104] fix: resolved bug on authorization string check

andrea-deri:
This PR contains several changes made on APIM policy XMLs in order to improve the security for the platform Authorization. Before this update, it was possible for someone to use a substring of the stored entity for pass the check of the platform Authorization fragment. This could cause security issues due to the fact that an attacker can use this exploit to use APIs on which it was not authorized.

#### Notes
Before applying this updates, it is necessary to delete all the _cached authorization_ (using Redis client) because this new logic will cause the existing one to be obsolete and no more usable.
To Apply this changes, it was executed the following command:
```
sh terraform.sh apply ENV \
--target=apim_api_authorizer_api_v1.azurerm_api_management_api_policy \
--target=azapi_resource.authorizer_fragment 
```

### List of changes
 - Updated API logic, adding escaping characters to the auth entity value
 - Updated fragment policy, adding escaping character evaluation to the object to be cached 

### Motivation and context
This changes are required in order to resolve the security issue previously described.

### Type of changes
- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/shared-app/api/authorizer/authorizer-check.xml
+++ b/src/domains/shared-app/api/authorizer/authorizer-check.xml
@@ -7,7 +7,7 @@
     <trace source="authorizer-check" severity="information">authorizer-check</trace>
     <!-- Saving information from request -->
     <set-variable name="domain" value="@((string)context.Variables["application_domain"])" /> <!-- ex. gpd -->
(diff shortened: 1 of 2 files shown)


pasqualespica (approved):
lgtm 👍

pasqualespica:
included in https://github.com/pagopa/pagopa-infra/pull/1179

PR #1162, opened 2023-07-19
fix(authRequest policy): changed where the <base/> policy is invoked …

Fernando-Granato:
### List of changes
Modified the point of invoke the base policy into **_auth_request.xml.tpl**. 🙂

### Motivation and context
The change was made to solve the problem of unconfigured cors in case of 401 http status code for invalide JWT. 👀

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pietro-tota (approved):

ciuffagianluca (approved):
LGTM!

PR #1179, opened 2023-07-21
[PAGOPA-1112] fix: resolved bug on locked internal state

andrea-deri:
This PR contains several changes made on APIM policy XMLs in order to resolve a bug on platform Authorizer. This bug, caused by a bad ending of Azure functions functioning during authorization caching, forced the `authorization_DOMAIN_in_progress` variable to remain in a dangling state where it cannot be automatically destroyed by policy. It was deleted only when the TTL comes to 0. This caused the execution to be locked due to the fact that the Authorizer policy fragment reads the dangling `authorization_DOMAIN_in_progress` flag and cannot perform nothing because the lock state impose to stop processing.
Before closing this PR, please merge [this one](https://github.com/pagopa/pagopa-infra/pull/1157) since the updates made here are forked from this pending PR. 

#### Notes
To Apply this changes, it was executed the following command:
```
sh terraform.sh apply ENV \
--target=apim_api_authorizer_api_v1.azurerm_api_management_api_policy \
--target=azapi_resource.authorizer_fragment 
```

### List of changes
 - Updated API logic, reducing the TTL value for the `authorization_DOMAIN_in_progress` variable
 - Updated fragment policy, reducing the TTL value for the `authorization_DOMAIN_in_progress` variable

### Motivation and context
This changes are required in order to resolve the bug previously described.

### Type of changes
- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/shared-app/api/authorizer/authorizer-check.xml
+++ b/src/domains/shared-app/api/authorizer/authorizer-check.xml
@@ -21,7 +21,7 @@
         <when condition="@(!"store".Equals(context.Variables["apim_cache_domain"]) && !context.Variables.ContainsKey("authorizer_domain_in_progress"))">
 
             <!-- Save state variable for locking purpose -->
-            <cache-store-value key="@("authorizer_" + context.Variables["domain"] + "_in_progress")" value="True" duration="86400" caching-type="internal" />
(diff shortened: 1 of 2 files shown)


pasqualespica (approved):
great

PR #1183, opened 2023-07-24
[Chk-1723] Create azure secret webhook

ciuffagianluca:
In order to trigger a github action we need a github token secret hub to store github token value

### List of changes

Add secrets in ecommerce domain

### Motivation and context

Add this secret to store github token value

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

Need to update value on the portal


giovanniberti (dismissed):

Fernando-Granato (dismissed):

ciuffagianluca:
This secret storage become unuseful since we use general purpose token for our scopes

PR #1204, opened 2023-07-31
[PAGOPA-1138] fix: resolved bug on auth store state's TTL

andrea-deri:
This PR contains the fix for a bug related to the platform Authorizer service. This bug is related to the TTL value associated to the store status' variable generated for domains. When an authorization was upserted in the CosmosDB, the function was triggered and the related authorization was updated. But on this event, also the status variable was updated and its life-cycle was extended. When its life-cycle is extended, the other authorization will expires before the store expiration. This means that for a certain amount of time the authorizations cannot be cached again and the user will obtain a wrong 403 code.
More details are present [in this Jira tasks](https://pagopa.atlassian.net/browse/PAGOPA-938).

### List of changes
 - Added conditional update for the store status variable

### Motivation and context
This PR permits to resolve the bug previously explained.

### Type of changes
- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/shared-app/.terraform.lock.hcl
+++ b/src/domains/shared-app/.terraform.lock.hcl
@@ -5,7 +5,6 @@ provider "registry.terraform.io/azure/azapi" {
   version     = "1.3.0"
   constraints = "1.3.0"
   hashes = [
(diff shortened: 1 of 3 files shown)


jacopocarlini (dismissed):
LGTM

pasqualespica (dismissed):

PR #1208, opened 2023-08-01
chore: Delete GPD-Core App Service [PAGOPA-1132]

AngeloCaporaso:
### List of changes

The following resources and modules were removed:
- `gpd_rgroup` (only terraform code)
- `gpd_service_plan` (destroyed)
- `gpd_snet` (A/A)
- `gpd_app_service` (A/A)
- `gpd_app_service_slot_staging` (A/A)
- `gpd_app_service_autoscale` (A/A)
- all variables related to the above resources

> Moreover, the resource `gpd_rgroup` was removed from the state but not destroyed as it was used and managed (created) in gps-common domain

`sh terraform.sh state weu-dev rm azurerm_resource_group.gpd_rgroup`

### Motivation and context

The GPD-Core App Service has been migrated to the AKS environment and is therefore no longer used.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
nice 👍

PR #1271, opened 2023-09-04
feat(jwt) added check jwt for api to retrieve card data

Fernando-Granato:
### List of changes
👉🏻 Added new resource `get_card_data_information` for execute jwt check for the api:
- GET /payment-methods/{id}/sessions/{sessionId}

👉🏻 Added new header into request **x-transaction-id-from-client** validate jwt

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


ciuffagianluca (approved):
well done!

infantesimone (approved):

PR #1290, opened 2023-09-08
feat: Add new nodo alerts [PAGOPA-1207]

AngeloCaporaso:
### Motivation and context

Add new nodo alerts targeted for each `primitiva` and sub-service.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
kudos @AngeloCaporaso

PR #1297, opened 2023-09-13
feat(nodo-app): Add response time alert for each operation [PAGOPA-1207] 

AngeloCaporaso:
### List of changes

- Add the option to customize the response time for each alert that refers to an operation;
- Update the trigger logic, which means that alerts are triggered when the following events occur: 2 thresholds exceeded in a 10-minute time window, calculated over a 5-minute interval.

### Motivation and context

For some operations having a response time greater than the one defined by default can be considered a normal behavior.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (dismissed):
LGTM great 👍

PR #1308, opened 2023-09-15
feat: Update nodo alerts [PAGOPA-1207] 

AngeloCaporaso:
### List of changes

- According to current behavior, custom `response-time` was updated for the following operation: `nodoInviaFlussoRendicontazione`, `nodoChiediNumeroAvviso`

- Availability query was updated

### Motivation and context

Improve nodo alerts according to current behavior. Goal:  reduce false positives.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/nodo-app/00_alert.tf
+++ b/src/domains/nodo-app/00_alert.tf
@@ -417,7 +417,6 @@ AzureDiagnostics
     Success=count(responseCode_d < 500)
     by bin(TimeGenerated, 5m)
 | extend availability=toreal(Success) / Total
-| where Total > 100
 | where availability < threshold
   QUERY
   )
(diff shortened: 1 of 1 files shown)


src/domains/nodo-app/00_alert.tf
```diff
@@ -412,6 +417,7 @@ AzureDiagnostics
     Success=count(responseCode_d < 500)
     by bin(TimeGenerated, 5m)
 | extend availability=toreal(Success) / Total
+| where Total > 100
```
pasqualespica: add let constant

pasqualespica (approved):

PR #1335, opened 2023-09-27
feat: Add nodo alerts on fault code [PAGOPA-1207]

AngeloCaporaso:
### List of changes

Two alerts on fault code was added:
- In the first one responses that return OK and those with one of these error codes ["PPT_PAGAMENTO_IN_CORSO","PPT_PAGAMENTO_DUPLICATO", "PPT_STAZIONE_INT_PA_TIMEOUT","PPT_ATTIVAZIONE_IN_CORSO", "PPT_ERRORE_EMESSO_DA_PAA"] are considered positive. The others are considered failure. The alert is triggered based on availability threshold.
- The second monitors PPT_ERRORE_EMESSO_DA_PAA error type only and is based on 2 thresholds: the availability theshold and the number of PSP impacted.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/nodo-app/00_alert_faultcode.tf
+++ b/src/domains/nodo-app/00_alert_faultcode.tf
@@ -4,6 +4,8 @@
 ## NB: In this case the fault code PPT_ERRORE_EMESSO_DA_PAA is considered as 'Success', because there is a dedicated alert for this type of error.
 ## Excluding this type of fault from OK group would complicate the definition of the general alert for fault codes like this.
 resource "azurerm_monitor_scheduled_query_rules_alert" "alert-fault-code-availability" {
+  count = var.env_short == "p" ? 1 : 0
+
   resource_group_name = "dashboards"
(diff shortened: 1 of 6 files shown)


src/domains/nodo-app/00_alert_faultcode.tf
```diff
@@ -0,0 +1,105 @@
...
+  frequency   = 5
+  time_window = 10
+  trigger {
+    operator  = "GreaterThanOrEqual"
+    threshold = 2
+  }
+}
+
+## If the number of fault codes of type PPT_ERRORE_EMESSO_DA_PAA increases, the alert is triggered. The alert takes into account how many PSPs are impacted.
+resource "azurerm_monitor_scheduled_query_rules_alert" "alert-fault-pa-error-availability" {
```
pasqualespica: for now we apply only first one to avoid false positive

pasqualespica (approved):
👍

PR #1349, opened 2023-10-02
feat(api-v2): open-api transaction-service

pietro-tota:
### List of changes
Add version 2 for eCommerce transactions-service api for new transaction creation (POST transactions)

### Motivation and context
Introduce transactions-service transaction activation api version 2.
This new version is required in order to add NPG orderId mandatory field used during transaction activation avoiding modifications to the v1 api version (as it's used by production Checkout version)

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


Fernando-Granato (approved):
🚀

src/domains/ecommerce-app/api/ecommerce-transactions-service/v2/_base_policy.xml.tpl
```diff
@@ -0,0 +1,17 @@
+<policies>
+    <inbound>
+      <set-header name="X-Client-Id" exists-action="delete" />
+      <!--> Add here some logic for value X-Client-Id header properly such as based on api key etc <-->
+      <base />
+      <set-backend-service base-url="https://${hostname}/pagopa-ecommerce-transactions-service/v2" />
```
infantesimone, line 6: is v2 required?
if we add the `v2` in the openapi file I think we can avoid it here

pietro-tota: Yes we can use both approach, I will change openapi descriptors :)

pietro-tota: I tried modifying open api path (adding preceding v2) but this way v2 prefix will be duplicated, one from api version set in api module and one from openapi path so I would leave this as is. What do you think?

PR #1370, opened 2023-10-10
Chk 1963 setup wallet mongodb

ciuffagianluca:
Create wallet resources for collection storage

### List of changes

Add configuration for database cosmodb and standard alert (as done for ecommerce)

### Motivation and context

This PR has the goal of providing the cosmodb resources and alert for wallet-service

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/wallet-common/03_cosmosdb.tf
```diff
@@ -74,3 +74,124 @@ resource "azurerm_cosmosdb_mongo_database" "wallet" {
...
+      indexes = [{
+          keys   = ["_id"]
+          unique = true
+        },
+        {
+          keys   = ["walletId", "eventType", "timestamp"]
+          unique = true
+        },
+        {
+          keys   = ["walletId"]
```
pietro-tota, line 116: this index should be already covered by the above one, so maybe it's useless. What do you think?

ciuffagianluca: We updated the indexes composition and definition. Thank you

src/domains/wallet-common/03_cosmosdb.tf
```diff
@@ -74,3 +74,120 @@ resource "azurerm_cosmosdb_mongo_database" "wallet" {
...
+
+  name                = each.value.name
+  resource_group_name = azurerm_resource_group.cosmosdb_wallet_rg[0].name
+
+  cosmosdb_mongo_account_name  = module.cosmosdb_account_mongodb[0].name
+  cosmosdb_mongo_database_name = azurerm_cosmosdb_mongo_database.wallet[0].name
+
+  indexes     = each.value.indexes
+  shard_key   = each.value.shard_key
+  lock_enable = var.env_short == "d" ? false : true
```
infantesimone, line 139: since we are refining the models/db, it is better if we enable the lock only in PROD

infantesimone (changes requested):

src/domains/wallet-common/03_cosmosdb.tf
ciuffagianluca: You're right!

infantesimone (approved):

PR #1381, opened 2023-10-13
feat: CHK-1995 add walletId at the end of onboarding

infantesimone:
### List of changes

- add `walletId` in cookie at the beginning of onboarding (`/webview/transactions/cc/verify`);
- add `walletId` at the end of onboarding of credit card (`/logout?walletId=123`)

### Motivation and context

The goal of this PR is to introduce a cookie at the beginning of the onboarding process in order to read at the end and return it together with the outcome, so `/logout?outcome=0&walletId=123`;

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/core/api/payment_manager_api/restapi-cd/v3/_base_policy.xml.tpl
```diff
@@ -36,6 +42,38 @@
           </set-header>
         </when>
       </choose>
+      <!-- cookie for walletId for new wallet backward compatible -->
+      <choose>
+          <when condition="@( context.Request.Url.Path.EndsWith("/webview/transactions/cc/verify") )">
+              <set-variable name="walletId" value="@(((string)context.Variables["requestBody"]).Split('&')[0].Split('=')[1])" />
```
pietro-tota, line 48: If I understood correctly, the intention here is to extract the first query parameter value for input request body. As for now exceptions can be throws for null input request body and for empty returned Split array.
I would rather add some check and split this in middle operations.
Something like:
```suggestion
              <set-variable name="walletId" value="@{
                string requestBody = ((string)context.Variables["requestBody"]);
                string parameters = requestBody!=null && requestBody.Split('?').Length >= 1 ? requestBody.Split('?')[1] : "";
                string[] queryParams = parameters != null ? queryParams = parameters.Split('&'): null;
                string[] firstQueryParam = queryParams != null && queryParams.Length >=1 ? queryParams[0].Split('=') : null;
                return firstQueryParam!=null && firstQueryParam.Length ==2 ? firstQueryParam[1]: "";
                }" />
```
What do you think? Obviously this ternary if operations can be exploded ad classic if statement too

pietro-tota: I was also exploring some C# utility function for URI parsing but I'm not sure that those are available in policy execution runtime

src/core/api/payment_manager_api/restapi-cd/v3/_base_policy.xml.tpl
```diff
@@ -36,6 +42,38 @@
...
+              <set-header name="Set-Cookie" exists-action="append">
+                  <value>@($"walletId={(string)context.Variables.GetValueOrDefault<string>("walletId","")}; Path=/pp-restapi-CD")</value>
+              </set-header>
+          </when>
+      </choose>
+      <choose>
+        <when condition="@( context.Request.Body != null && context.Request.Url.Path.Contains("/logout") )">
```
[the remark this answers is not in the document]
infantesimone: right

pietro-tota (approved):
🚀

PR #1385, opened 2023-10-13
[PRDP-167] feat: add cdn config in receipts

alessio-cialini:
### List of changes

- Introducing CDN configurations in receipts domain

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica:
this PR is no longer necessary, after check , [we already have one](https://github.com/pagopa/pagopa-infra/blob/306b3d10936096590b6bfeea2efcc5486101f07b/src/core/assets_cdn_platform.tf#L23)

PR #1399, opened 2023-10-18
feat(sops): introduce sops for managing secrets

pietro-tota:
### List of changes
Add quality improvements secrets domain.
This domain is used to store sops encoded secrets that will be applied in each environment.

### Motivation and context
This modification allow secrets to be securely versioned

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/qi-common/02_security.tf
```diff
@@ -41,7 +41,7 @@ resource "azurerm_key_vault_access_policy" "adgroup_developers_policy" {
   tenant_id = data.azurerm_client_config.current.tenant_id
   object_id = data.azuread_group.adgroup_developers.object_id
 
-  key_permissions     = ["Get", "List", "Update", "Create", "Import", "Delete", ]
+  key_permissions     = ["Get", "List", "Update", "Create", "Import", "Delete", "Encrypt", "Decrypt"]
```
pietro-tota, line 44: @pasqualespica here we add permissions also for Decrypt but maybe it should be lowered to Encrypt only. I take those permissions from [here](https://github.com/pagopa/pagopa-infra/blob/3dfcd939c2927a8e58cc0851cc79906fb0b02b74/src/domains/receipts-secrets/02_security.tf#L34-L49)
```suggestion
  key_permissions     = ["Get", "List", "Update", "Create", "Import", "Delete", "Encrypt"]
```
What do you think?

pasqualespica (approved):
only 1 thing maybe terraform `1.6.0` was used.
With this version vs `.terraform-version` i obtain this [error](https://github.com/hashicorp/terraform/issues/33770)

I resolved setting `1.6.2.` ( for ex) into `.terraform-version`

cc @infantesimone @pietro-tota

pasqualespica (approved):
👍

PR #1413, opened 2023-10-23
[PRDP-176] feat: Added alert for receipt generator

gioelemella:
### List of changes

- Added alert for ReceiptGenerator in case of missing property in biz-event
- Added alert for ReceiptGenerator in case of PDF Engine invocation KO
- Added alert for ReceiptGenerator in case of error on saving PDF in Blob Storage

### Motivation and context
Added alert

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (approved):
great work

PR #1444, opened 2023-10-30
Changing secret value dev QI

FedericoRuzzier:
### List of changes
Modifying dev secret value for JSM

### Motivation and context
In order to avoid opening real tickets when developing

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pietro-tota (approved):
LGTM

PR #1457, opened 2023-11-06
chore: Update path in GPD PN Integration API [PAGOPA-1338]

AngeloCaporaso:
### List of changes

path update in `GPD PN Integration` API 

### Motivation and context

PN integration API was updated to exclude it from deny-paths

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (approved):
👍

PR #1478, opened 2023-11-09
[PRDP-196] feat: introducing tokenizer client error alerts

alessio-cialini:
### List of changes

- Introducing error alerts for services in the receipt domain, regarding the errors in the usage of the PDV Tokenizer Client

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
👍

PR #1490, opened 2023-11-14
chore(nodo-alert): Update fault-code alert

AngeloCaporaso:
### List of changes

- Update the range in which the threshold is 80;
- Update on cases where fault-code almost certainly isn't related to pagoPA fault;

### Motivation and context

The purpose is to reduce false positives related to this alert.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (dismissed):
LGTM 👍

pasqualespica (approved):

PR #1504, opened 2023-11-16
[CHK-2135] add test for different type of credit card

ciuffagianluca:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


ciuffagianluca:
No need for this pr. The intent is to have all kind of available wallet from the same wallet token

PR #1551, opened 2023-11-30
[CHK-2180] feat: enable Opsgenie alerts for eCommerce

giovanniberti:
### List of changes

 * moved Opsgenie action group to `ecommerce-commons`
 * added Opsgenie action to all alerts for eCommerce

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


infantesimone:
if it's not too out of scope, do we move all alerts in `common`? or there is a reason for having alerts in `common` and some in `app`?

giovanniberti:
Sure, I think some alerts are inside `ecommerce-commons` for historical reasons

PR #1575, opened 2023-12-12
[CHK-2228] feat: redefine npg notification url

ciuffagianluca:
Edit npg notification url spec and policy

### List of changes

- Npg notification url (for all ecommerce transactions) now has `sessionToken` queryParam.
- Added policy jwt validation on sessionToken value.
- Match `orderId` value and get value about `transactionId` and `paymentMethodId`

### Motivation and context

Secure npg notification url
Make it compliant with logged user payment transaction

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl
+++ b/src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl
@@ -3,7 +3,7 @@
         <base />
         <!-- start validation policy -->
         <set-variable name="orderId" value="@(context.Request.MatchedParameters["orderId"])" />
-        <validate-jwt query-param-name="sessionToken" failed-validation-httpcode="401" failed-validation-error-message="Unauthorized" require-expiration-time="true" require-signed-tokens="true" output-token-variable-name="jwtToken">
+        <validate-jwt query-parameter-name="sessionToken" failed-validation-httpcode="401" failed-validation-error-message="Unauthorized" require-expiration-time="true" require-signed-tokens="true" output-token-variable-name="jwtToken">
         <issuer-signing-keys>
             <key>{{npg-notification-jwt-secret}}</key>
         </issuer-signing-keys>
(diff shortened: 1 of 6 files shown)


src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl
```diff
@@ -1,11 +1,43 @@
 <policies>
     <inbound>
         <base />
-        <!-- start policy variables -->
+        <!-- start validation policy -->
+        <set-variable name="orderId" value="@(context.Request.MatchedParameters["orderId"])" />
+        <validate-jwt query-param-name="sessionToken" failed-validation-httpcode="401" failed-validation-error-message="Unauthorized" require-expiration-time="true" require-signed-tokens="true" output-token-variable-name="jwtToken">
```
infantesimone, line 6: Where is the token generated? because we could use a different key to generate the tokens for the `notificationUrl` than those used for the checkout/ecommerce payment flow.

infantesimone: `ecommerce-jwt-signing-key` is used for payment-flow

ciuffagianluca: The token is generated in payment method or transactions service, i guess using the same key. 
I agree with you that the key for sure is different from the key used in checkout/ecommerce payment flow.
But do you agree with me that both `jwtToken` that have to passed to the `esito url ` generated from `paymentMethod` (for guest payment) and from `transaction service` (for logged payment) must be **signed** with the **same key**?

infantesimone: yep

infantesimone (dismissed):
just update `ecommerce-common` readme with the new secret

src/domains/ecommerce-common/02_security.tf
```diff
@@ -441,6 +441,19 @@ resource "azurerm_key_vault_secret" "wallet-api-key" {
...
 
+  lifecycle {
+    ignore_changes = [
+      value,
+    ]
+  }
+}
+
+
+resource "azurerm_key_vault_secret" "npg_notification_signing_key" {
```
ciuffagianluca, line 452: @infantesimone isn't this one the new secret you refer to in your comment?

infantesimone (approved):

PR #1580, opened 2023-12-13
feat: added Selfcare Backoffice (BE and FE) for version info

andrea-deri:
This PR contains the changes applied on pagoPA's Status Page in order to add the info status for SelfCare Backoffice frontend and backend application.
Also, a new procedure for getting the version info for APIConfig frontend application was added and will be integrated as soon as possible from this repository.

### List of changes
 - Added version info for Selfcare Backoffice backend
 - Added version info for Selfcare Backoffice frontend
 - Added version info for APIConfig frontend

### Type of changes
- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/shared-app/04_apim_statuspage.tf
+++ b/src/domains/shared-app/04_apim_statuspage.tf
@@ -1,3 +1,13 @@
+data "azurerm_storage_account" "pagopa_selfcare_fe_sa" {
+  name                = replace("${var.prefix}-${var.env_short}-selfcare-sa", "-", "")
+  resource_group_name = format("%s-%s-fe-rg", var.prefix, var.env_short)
+}
+
+data "azurerm_storage_account" "pagopa_apiconfig_fe_sa" {
+  name                = replace("${var.prefix}-${var.env_short}-${var.location_short}-apiconfig-sa", "-", "")
+  resource_group_name = format("%s-%s-api-config-rg", var.prefix, var.env_short)
+}
+
 ##############
 ## Products ##
 ##############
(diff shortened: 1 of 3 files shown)


src/domains/shared-app/04_apim_statuspage.tf
```diff
@@ -31,7 +31,9 @@ locals {
     subscription_required = false
     service_url           = null
   }
-  aks_path = var.env == "prod" ? "weuprod.%s.internal.platform.pagopa.it" : "weu${var.env}.%s.internal.${var.env}.platform.pagopa.it"
+  aks_path           = var.env == "prod" ? "weuprod.%s.internal.platform.pagopa.it" : "weu${var.env}.%s.internal.${var.env}.platform.pagopa.it"
+  fe_backoffice_path = "pagopa${var.env_short}selfcaresa.z6.web.core.windows.net/ui/version.json"
```
pasqualespica, line 35: why not this https://github.com/pagopa/terraform-azurerm-v3/blob/a788d9cd14f2f1157b72aa0d011f75ad56b12b0b/storage_account/outputs.tf#L24 ? 
image

src/domains/shared-app/api/status-page-service/v1/_base_policy.xml
```diff
@@ -29,11 +29,17 @@
...
           <choose>
-            <when condition="@(context.Variables["baseUrl"] != null)">
-            <set-backend-service base-url="@{
-                          var services = JObject.Parse((string)context.Variables["services"]);
-                          return "https://" + (string)services[context.Variables["product"]];
-            }" />
+            <when condition="@(context.Variables["baseUrl"] != null && ((string)context.Variables["product"]).EndsWith("-fe"))">
+              <set-backend-service base-url="@{
+                                var services = JObject.Parse((string)context.Variables["services"]);
+                                return "https://" + (string)services[context.Variables["product"]] + "?";
```
pasqualespica, line 35: if th purpose is to add `?` or not., why not use `if` in to `{ ...}` code ?

src/domains/shared-app/04_apim_statuspage.tf
andrea-deri: Good idea! This will be changed as you describes, both for SelfCare BO Frontend and for ApiConfig Frontend URLs.

src/domains/shared-app/api/status-page-service/v1/_base_policy.xml
andrea-deri: The '?' character is needed because seems that this policy was made in order to replace only the base URL. 
This character force to pass the section after the base path as a query parameter (that will be ignored during the redirect). I don't like it either but at the moment seems a way to not refactor the whole policy. Feel free to suggest a better way to do this!

jacopocarlini (approved):

PR #1592, opened 2023-12-15
fix: Verify KO

aomegax:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/nodo-common/03_cosmosdb_nodo_verifyko.tf
```diff
@@ -1,9 +1,16 @@
...
+  tags = var.tags
+}
+
 module "cosmosdb_account_nodo_verifyko" {
   source              = "git::https://github.com/pagopa/terraform-azurerm-v3.git//cosmosdb_account?ref=v6.7.0"
   domain              = var.domain
   name                = "${local.project}-verifyko-cosmos-account"
   location            = var.location
-  resource_group_name = azurerm_resource_group.db_rg.name
+  resource_group_name = azurerm_resource_group.nodo_verifyko_to_datastore_rg.name
```
mamari90, line 13: @aomegax check the verifyko_cosmos_nosql_db_params variable for prod env; the new account must be zone redundant in the primary region
also check the max throughput, is it necessary to have 20000 RU (that will be doubled in a multi-region configuration like this one)?

mamari90: note that adding zone redundancy forces re-creation of the resource

src/domains/nodo-common/03_storage_account_nodo_verifyko.tf
```diff
@@ -1,7 +1,3 @@
-data "azurerm_resource_group" "nodo_verifyko_to_datastore_rg" {
-  name = format("%s-verifyko-to-datastore-rg", local.project)
-}
-
 module "nodo_verifyko_storage_account" {
```
mamari90, line 1: @aomegax if this storage account requires to be backed up (in prod env), you can add the related configuration alongside delete_retention and versioning_enabled:

```
blob_change_feed_enabled             = var.afm_storage_params.enable_backup
  blob_change_feed_retention_in_days   = var.afm_storage_params.enable_backup ? var.afm_storage_params.backup_retention_days + 1 : null
  blob_container_delete_retention_days = var.afm_storage_params.backup_retention_days
  blob_storage_policy = {
    enable_immutability_policy = false
    blob_restore_policy_days   = var.afm_storage_params.backup_retention_days
  }
  ```
  
  check `module "afm_storage"` for an example

PR #1599, opened 2023-12-18
chore: Enable pdf receipt 2 all IO users 🚀 🤞 

pasqualespica:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica:
all done into other PR

PR #1632, opened 2024-01-11
[CHK 2309] feat added get all payment method

Fernando-Granato:
### List of changes

- Added the endpoint for retrieving all payment methods into the openapi **_ecommerce-for-io_**.
          The header x-client-Id io is automatically set in the _base_policy of the api group 👉🏻 https://github.com/pagopa/pagopa-infra/blob/5c799513a1049b4eb5d16c149b987b077651adc7/src/domains/ecommerce-app/api/ecommerce-io/v1/_base_policy.xml.tpl#L37
- Furthermore, the payment-method and checkout open APIs have been updated with the addition of the enum PaymentMethodManagementType in the **PaymentMethodResponse** response and PaymentMethodRequest DTO

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-checkout/v1/_openapi.json.tpl
```diff
@@ -1792,6 +1792,10 @@
             "items": {
               "$ref": "#/components/schemas/Range"
             }
+          },
+          "isOnboardable": {
```
pietro-tota, line 1796: I would made this field mandatory in payment method response

Fernando-Granato: And if not present on the collection we set it to false by default ?

pietro-tota: Yes, I would handle the field absence into payment method logic (or make isOnboardable field mandatory for all payment method documents)

Fernando-Granato: Fixed with 09f90ad8f46c088662fd63be94ccea3e85947e3f

pietro-tota (approved):
🚀

pietro-tota (approved):
🚀

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_openapi.json.tpl
```diff
@@ -1800,7 +1803,8 @@
           "description",
           "status",
           "paymentTypeCode",
-          "ranges"
+          "ranges",
+          "methodManagement"
```
infantesimone, line 1807: a required field for the api currently used in prod by checkout, is this PR backwards compatible? does it require a definition of a release strategy?

infantesimone (changes requested):

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_openapi.json.tpl
Fernando-Granato: Momentarily we could avoid updating the openapi of **ecommerce for checkout** by ensuring backward compatibility. What do you think ?

infantesimone (approved):

PR #1656, opened 2024-01-17
[CHK-2348] fix(walletType): remove enum from wallet type and use free string

ciuffagianluca:
remove enum from wallet type and use free string

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/wallet-app/api/payment-wallet/v1/_openapi.json.tpl
```diff
@@ -669,7 +615,7 @@
               "expiryDate": {
                 "type": "string",
                 "description": "Credit card expiry date. The date format is `YYYYMM`",
-                "pattern": "^\\d{6}$",
+                "pattern": "^\\\\d{6}$",
```
ciuffagianluca, line 618:
```suggestion
                "pattern": "^[0-9]{6}$",
```

infantesimone (approved):

PR #1677, opened 2024-01-25
NOD-617

lorenzo-catalano:
[CONFIG CACHE] Creazione script liquibase per il nodo con la nuova tabella per la cache

### List of changes
Added liquiabse scripts for nodo config cache only

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


aomegax:
not useful

PR #1711, opened 2024-02-02
PAGOPA-1500 adding new archive sa

FedericoRuzzier:
### List of changes
Adding new sa for the payment position archive, removing unused sa for the receipt of gpd

### Motivation and context
This addition was needed in order to create a storage unit to archive the old payment position

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/gps-common/03_storage_account.tf
```diff
@@ -3,37 +3,44 @@
 ##################
 
 # storage
-module "payments_receipt_sa" {
+module "gpd_archive_sa" {
   source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//storage_account?ref=v7.18.0"
 
-  name                            = replace(format("%s-payments-sa", local.project), "-", "")
+  name                            = replace(format("%s-archive-sa", local.project), "-", "")
```
AngeloCaporaso, line 9: so the name looks like `<prefix><env><location>gpsarchivesa`?
maybe it's better `<prefix><env><location>gpdarchivesa`

src/domains/gps-common/03_storage_account.tf
```diff
@@ -3,37 +3,44 @@
 ##################
 
 # storage
-module "payments_receipt_sa" {
```
AngeloCaporaso, line 6: Has it been removed because it was not being used? Is this the storage migrated to cosmos table?

src/domains/gps-common/03_storage_account.tf
```diff
@@ -3,37 +3,44 @@
...
+module "gpd_archive_sa" {
   source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//storage_account?ref=v7.18.0"
 
-  name                            = replace(format("%s-payments-sa", local.project), "-", "")
+  name                            = replace(format("%s-archive-sa", local.project), "-", "")
   account_kind                    = "StorageV2"
   account_tier                    = "Standard"
   account_replication_type        = var.storage_account_replication_type
```
AngeloCaporaso, line 12: there isn't a dedicated variables for `storage_account_replication_type`. If doesn't exist archive policy It could be `GZRS`

AngeloCaporaso (changes requested):
Good work, I left some comments to improve the solution

src/domains/gps-common/03_storage_account.tf
pasqualespica: yes we shall removed it

PR #1727, opened 2024-02-07
[CHK-2448] feat(closePayment): add field variant for `additionalPaymentInformation` to encode Checkout Redirect payments

giovanniberti:
### List of changes

 * ported spec to OpenAPI v3
 * added `additionalPaymentInformation` field variant for Checkout Redirect

### Motivation and context

This change allows the pagoPA platform to correctly collect information about payments done through Checkout Redirect

[Check out the updated API here](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pagopa/pagopa-infra/CHK-2448-redirect-additionalpaymentinformations/src/core/api/nodopagamenti_api/nodoPerPM/v2/_openapi.json.tpl)

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information
NOTE: apim resource definition still refers swagger file (that have not been upgraded) since changing from swagger to openapi will require a resource re-creation and will be done in a separate PR

---


src/core/api/nodopagamenti_api/nodoPerPM/v2/_swagger.json.tpl
```diff
@@ -1,20 +1,15 @@
 {
-  "swagger": "2.0",
+  "openapi": "3.0.1",
```
pietro-tota, line 2: this file can be renamed to `_openapi.json.tpl`, what do you think?

src/core/api/nodopagamenti_api/nodoPerPM/v2/_swagger.json.tpl
```diff
@@ -26,350 +21,475 @@
...
+            "description": "Redirect payment method name"
+          },
+          "additionalPaymentInformations": {
+            "type": "object",
+            "description": "Additional payment data pertaining to Redirect-type payments",
+            "properties": {
+              "tipoVersamento": {
+                "type": "string",
+                "description": "Specific Checkout Redirect gateway"
+              },
```
pietro-tota, line 198: are you sure this field has to be added for redirect close payments? why is in italian?

giovanniberti: Copy-paste gone wrong, removing it! Thanks 👍

giovanniberti: Fixed with 0f17d1c5

src/core/api/nodopagamenti_api/nodoPerPM/v2/_swagger.json.tpl
giovanniberti: Fixed with 07563725

pietro-tota (dismissed):

infantesimone:
It's correct and interesting to move to OpenAPI, but this requires the destruction and recreation of the APIs (downtime in production). Therefore, if we want to perform this refactoring, we need to plan it thoroughly with our colleagues @pagopa/pagopa-team-core .

src/core/api/nodopagamenti_api/nodoPerPM/v2/_openapi.json.tpl
```diff
@@ -0,0 +1,455 @@
...
+          "error"
+        ],
+        "type": "object",
+        "properties": {
+          "error": {
+            "type": "string",
+            "example": "error message"
+          }
+        }
+      },
```
pietro-tota, line 126: please fix the Error response body as done in this [PR](https://github.com/pagopa/pagopa-infra/pull/1731/files)

infantesimone:
Is it possible to manage additionalPaymentInformation with Swagger 2.0?

giovanniberti:
> Is it possible to manage additionalPaymentInformation with Swagger 2.0?

The problem is that [Swagger 2.0 doesn't support `oneOf`](https://stackoverflow.com/questions/70112647/equivalent-use-of-oneof-from-openapi-3-in-openapi-2), which is critical to have a descriptive spec (as well as having type safety in the generated implementations).

In the meantime a possibility could be to maintain both the Swagger and the OpenAPI specs so that clients could point to whatever spec they choose

src/core/api/nodopagamenti_api/nodoPerPM/v2/_openapi.json.tpl
giovanniberti: Fixed in 5e75845b

giovanniberti:
ping :)

pietro-tota (approved):

PR #1728, opened 2024-02-07
feat: Added synthetic monitoring module

mamari90:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/synthetic-monitoring/01_analytics_workspace.tf
+++ b/src/synthetic-monitoring/01_analytics_workspace.tf
@@ -1,0 +1,28 @@
+resource "azurerm_resource_group" "synthetic_rg" {
+  location = var.location
+  name     = "${local.project}-rg"
+}
+
+
+resource "azurerm_log_analytics_workspace" "log_analytics_workspace" {
+  name                = "${local.project}-law"
+  location            = azurerm_resource_group.synthetic_rg.location
+  resource_group_name = azurerm_resource_group.synthetic_rg.name
+  sku                 = var.law_sku
+  retention_in_days   = var.law_retention_in_days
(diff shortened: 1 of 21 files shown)


src/synthetic-monitoring/01_monitoring.tf
```diff
@@ -0,0 +1,56 @@
...
+    table_private_dns_zone_id = var.use_private_endpoint ? data.azurerm_private_dns_zone.storage_account_table.id : null
+    replication_type          = var.storage_account_replication_type
+  }
+
+  private_endpoint_subnet_id = var.use_private_endpoint ? data.azurerm_subnet.private_endpoint_subnet[0].id : null
+
+  tags = var.tags
+
+  self_alert_enabled = false
+  monitoring_configuration_encoded = jsonencode([
```
diegolagospagopa, line 40: this values is better to put into variables files

diegolagospagopa (changes requested):

src/synthetic-monitoring/01_analytics_workspace.tf
```diff
@@ -0,0 +1,28 @@
...
+  resource_group_name = azurerm_resource_group.synthetic_rg.name
+  sku                 = var.law_sku
+  retention_in_days   = var.law_retention_in_days
+  daily_quota_gb      = var.law_daily_quota_gb
+
+  tags = var.tags
+}
+
+# Application insights
+resource "azurerm_application_insights" "application_insights" {
```
diegolagospagopa, line 19: i think is better to create a generic monitoring_tools or something similar, to be able to use one for all

diegolagospagopa (changes requested):

PR #1771, opened 2024-02-27
[CHK-2510] authorization outcome waiting queue

ciuffagianluca:
Rename authorization requested retry event queue to authorization outcome waiting queue

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/ecommerce-common/03_storage.tf
+++ b/src/domains/ecommerce-common/03_storage.tf
@@ -115,8 +115,8 @@ resource "azurerm_storage_queue" "transactions_authorization_requested_queue" {
   storage_account_name = module.ecommerce_storage_transient.name
 }
 
-resource "azurerm_storage_queue" "transactions_authorization_requested_retry_queue" {
-  name                 = "${local.project}-transaction-auth-requested-retry-queue"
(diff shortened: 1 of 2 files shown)


src/domains/ecommerce-common/03_storage.tf
```diff
@@ -115,8 +115,8 @@ resource "azurerm_storage_queue" "transactions_authorization_requested_queue" {
   storage_account_name = module.ecommerce_storage_transient.name
 }
 
-resource "azurerm_storage_queue" "transactions_authorization_requested_retry_queue" {
-  name                 = "${local.project}-transaction-auth-requested-retry-queue"
+resource "azurerm_storage_queue" "transactions_authorization_outcome_waiting_queue" {
```
pietro-tota, line 118: maybe here we have to duplicate queue leaving the old one so that first apply will only add the new queue then proceed with modules deploy and remove the old queue in a separate pr. What do you think?

ciuffagianluca: This queue is not referenced in any services deployed in production. We can proceed to replace it directly. Anyway we can evaluate a gradual approach to avoid apply failure due to lock

infantesimone (approved):

PR #1802, opened 2024-03-06
[CHK-2560] wallet delete vs pm policy

ciuffagianluca:
Add policy to delete wallet vs pm
Add policy to invoke delete policy vs new wallet

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/wallet-app/api/payment-wallet/v1/_delete_wallet_with_pm.xml.tpl
```diff
@@ -0,0 +1,103 @@
...
+            </when>
+        </choose>
+        <set-variable name="pmSession" value="@(((IResponse)context.Variables["pm-session-body"]).Body.As<JObject>())" />
+        <!-- START delete wallet -->
+        <set-variable name="idWalletPM" value="@{
+            string walletId = context.Request.MatchedParameters["walletId"];
+            return Convert.ToInt32("0x" + walletId.Substring(walletId.Length - 12), 16);
+        }" />
+        <send-request ignore-error="false" timeout="10" response-variable-name="pmWalletDeleteResponse">
+            <set-url>@($"{{pm-host}}/pp-restapi-CD/v1/wallet/{(string)context.Variables["idWalletPM"]}")</set-url>
```
giovanniberti, line 33: Just to check: there is also a `DELETE /pp-restapi-CD/v1/wallet/{walletId}/{pspId}`, are we sure this works for us?

giovanniberti (approved):

src/domains/wallet-app/api/payment-wallet/v1/_delete_wallet_with_pm.xml.tpl
ciuffagianluca: Didn't try directly in policy, but i tried it via postman and it works. To be honest, I didn't know the difference, but at first sight I'm not so confident in this one, just for the problem to retrieve the `pspId` param. Anyway, maybe @infantesimone could help us to choice which is the API to invoke.

infantesimone: the API used in PR is fine, the other API is not used (perhaps it was done for a refactoring with the introduction of paypal...)

infantesimone (dismissed):

pietro-tota (approved):

infantesimone (approved):

PR #1823, opened 2024-03-13
gpd-core-swagger-update

jacopocarlini:
gpd-core-swagger-update


AngeloCaporaso:
only a test, I will close PR

PR #1853, opened 2024-03-25
feat: update checkout cdn configuration to handle NPG

infantesimone:
### List of changes

- add `https://assets.cdn.platform.pagopa.it` in `checkout` CSP to load new logo;
- add nexi domain in script-src to load NPG sdk.
- add new delivery rule to allow NPG iframe load css exposed from `checkout` CDN
### Motivation and context

The goal of this PR is to update checkout CDN configuration in order to handle NPG integration.
In addition to updating the CSPs, a new delivery rule has been created to overwrite headers in the response when CSS is loaded from NPG iframes. Currently, there were CORS issues for some of these resources:
Screenshot 2024-03-25 alle 19 49 46

with this delivery rule:
Screenshot 2024-03-25 alle 19 48 50

more details for the delivery rule https://learn.microsoft.com/en-us/azure/cdn/cdn-cors#azure-cdn-standard-profiles 

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [n] No

### Other information

---


src/domains/checkout-app/05_checkout_fe.tf
```diff
@@ -69,12 +69,12 @@ module "checkout_cdn" {
...
         name   = "Content-Security-Policy-Report-Only"
-        value  = "img-src 'self' https://cdn.cookielaw.org https://acardste.vaservices.eu:* https://wisp2.pagopa.gov.it https://assets.cdn.io.italia.it www.gstatic.com/recaptcha data:;"
+        value  = "img-src 'self' https://cdn.cookielaw.org https://acardste.vaservices.eu:* https://wisp2.pagopa.gov.it https://assets.cdn.io.italia.it www.gstatic.com/recaptcha data: https://assets.cdn.platform.pagopa.it;"
       },
       {
         action = "Append"
         name   = "Content-Security-Policy-Report-Only"
-        value  = "script-src 'self' 'unsafe-inline' https://www.google.com https://www.gstatic.com https://cdn.cookielaw.org https://geolocation.onetrust.com https://www.recaptcha.net https://recaptcha.net https://www.gstatic.com/recaptcha/ https://www.gstatic.cn/recaptcha/;"
+        value  = "script-src 'self' 'unsafe-inline' https://www.google.com https://www.gstatic.com https://cdn.cookielaw.org https://geolocation.onetrust.com https://www.recaptcha.net https://recaptcha.net https://www.gstatic.com/recaptcha/ https://www.gstatic.cn/recaptcha/ https://xpay.nexigroup.com https://stg-ta.nexigroup.com ;"
```
pietro-tota, line 77: can we split this value based on the env so that `https://stg-ta.nexigroup.com` is add only for dev and uat envs and not for prod ones? maybe replicating the same approach applied below.
What do you think?

pietro-tota (approved):
Well done, just a consideration :)

src/domains/checkout-app/05_checkout_fe.tf
infantesimone: good point

infantesimone: done

PR #1858, opened 2024-03-26
CHK-2639: cardid4 notify wallet

infantesimone:
### Motivation and context

The goal of this PR is to send the `additionalData.cardId4` (NPG request) to the `wallet-service`, so as to be able to check whether the card has already onboarded.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


ciuffagianluca (approved):

src/domains/wallet-app/api/npg-notifications/v1/_base_policy.xml.tpl
```diff
@@ -51,11 +51,19 @@
                 details = new JObject();
                 details["type"] = "PAYPAL";
                 details["maskedEmail"] = (string)operation["paymentInstrumentInfo"];
+            } else if(paymentCircuit == "CARD"){
+                details = new JObject();
+                details["type"] = "CARD";
+                if(additionalData.Type != JTokenType.Null){
+                    details["paymentInstrumentGatewayId"] = (string)((JObject)additionalData)["cardId4"];
```
pietro-tota, line 58: maybe you can move up check performed in line 62 above and avoid repeating the `if(additionalData.Type != JTokenType.Null){` multiple times, assigning a local variables to be used here. What do you think?

pietro-tota (approved):

PR #1895, opened 2024-04-04
refactor: CHK-2645 Update ecommerce for checkout auth request policy to zero downtime pgs->npg

infantesimone:
### List of changes

- Update ecommerce for checkout auth request policy to zero downtime pgs->npg

### Motivation and context

The goal of this PR is to refactor the policy for selecting the `pgsId` for the `checkout` flow with `eCommerce` to ensure zero downtime during the migration from `eCommerce` with `PGS` to `eCommerce` with `NPG`. So, the selection of the `pgsId` is based on the value of the `detailsType`, where `"card"` indicates a `PGS` flow, while `"cards"` indicates a `NPG` flow.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
+++ b/src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
@@ -21,7 +21,6 @@
         <!-- START set pgsId -->
         <set-variable name="XPAYPspsList" value="${ecommerce_xpay_psps_list}"/>
         <set-variable name="VPOSPspsList" value="${ecommerce_vpos_psps_list}"/>
-        <set-variable name="NPGPspsList" value="${ecommerce_npg_psps_list}"/>
         <set-variable name="requestBody" value="@((JObject)context.Request.Body.As<JObject>(true))" />
 
         <set-variable name="pspId" value="@(((string)((JObject)context.Variables["requestBody"])["pspId"]))"/>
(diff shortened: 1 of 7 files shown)


src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
```diff
@@ -18,26 +18,47 @@
...
+
+                    pgsId = "XPAY";
+                } else if (vposList.Contains(pspId)) {
+
+                    pgsId = "VPOS";
+                }
+
+            // cards -> ecommerce with NPG request
+
+            } else if ( detailType == "cards" && npgList.Contains(pspId)){
```
pietro-tota, line 55: As for now we have also `wallet`, `apm` and `redirect` detailTypes, the first two used for onboarded/guest payments, the last for redirect payment flows.

Since the gateways will be 3 as for now (two in the future when pgs will be shutdown) what about changing this else if condition to check against `redirect` detailType (and set pgsId to redirect) and then in the else branch assume that it's an NPG transaction?

Maybe we can also add a psp id based filtering for redirect transactions, what do you think?

ciuffagianluca (approved):

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_auth_request.xml.tpl
infantesimone: redirect added :-) thanks!

pietro-tota (dismissed):

pietro-tota (approved):

PR #1906, opened 2024-04-09
[CHK-2664] refactor: check and add response body in case of invalid token

giovanniberti:
### List of changes

 * eCommerce API for IO: add response body in case of invalid token
 * wallet: forward 401 http status code when checking token

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


infantesimone:
well done :-)

infantesimone (approved):

PR #1911, opened 2024-04-09
[CHK-2731] Add wallet storage connection string on ecommerce and wallet kv

petretiandrea:
Adds storage queue connection string to eCommerce and Wallet Keyvault. From eCommerce side is used to produce last wallet usage events and from wallet side is used to consume these events.

### List of changes
- add connection string to eCommerce kv
- add connection string to Wallet kv

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-common/02_security.tf
```diff
@@ -539,6 +539,18 @@ resource "azurerm_key_vault_secret" "npg_mybank_psp_keys" {
...
+  lifecycle {
+    ignore_changes = [
+      value,
+    ]
+  }
+}
+
+resource "azurerm_key_vault_secret" "wallet_storage_connection_string" {
+  name         = "wallet-storage-connection-string"
+  value        = "<TO UPDATE MANUALLY ON PORTAL>"
```
infantesimone, line 551: you could use `primary_connection_string` of [module output](https://github.com/pagopa/terraform-azurerm-v3/blob/97e011d8a6bebee22aaad8e02a40700fd7fca661/storage_account/outputs.tf#L9C9-L9C34) .

petretiandrea: The wallet storage is created on wallet domain side, so the storage connection string secret in `wallet-common` now reference the output of `wallet_storage` module. The same connection string is needed on eCommerce domain side, which cannot reference directly the storage module (unlike to reference it through data or something else), we can consider the connection string as "external secret" from eCommerce domain? (something like an api key) or create a reference via "data"?

infantesimone (approved):

PR #1956, opened 2024-04-18
feat(NPG monitoring): alert for slow or KO order/build api calls

pietro-tota:
### List of changes
Add alert that will be triggered when more than 10 POST `payment-methdods/{id}/sessions` requests are detected that have received a KO response or with a duration more than 2 seconds 

### Motivation and context
This api is used from Checkout f.e. to retrieve the card data fields for build the "insert card data" page.
A KO received here or slow api response times will affect the user experience causing the impossibility to perform a transaction payment

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/ecommerce-common/00_alerts.tf
+++ b/src/domains/ecommerce-common/00_alerts.tf
@@ -134,3 +134,37 @@ AzureDiagnostics
     threshold = 60 // NDP for each requests does 5 retries with a 2 minutes interval after a 10 minutes delay
   }
 }
+
+
+# eCommerce NPG monitoring: KO or slow payment methods start session api call (order/build retrieve card form fields)
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_payment_methods_start_session_alert" {
+  count = var.env_short == "p" ? 1 : 0
+
+  name                = "ecommerce-payment-methods-start-session-alert"
+  resource_group_name = azurerm_resource_group.rg_ecommerce_alerts[0].name
+  location            = var.location
+
+  action {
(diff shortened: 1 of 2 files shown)


giovanniberti (dismissed):

src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -134,3 +134,37 @@ AzureDiagnostics
...
+# eCommerce NPG monitoring: KO or slow payment methods start session api call (order/build retrieve card form fields)
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_payment_methods_start_session_alert" {
+  count = var.env_short == "p" ? 1 : 0
+
+  name                = "ecommerce-payment-methods-start-session-alert"
+  resource_group_name = azurerm_resource_group.rg_ecommerce_alerts[0].name
+  location            = var.location
+
+  action {
+    action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id, azurerm_monitor_action_group.ecommerce_opsgenie[0].id]
```
infantesimone, line 148: We should add alerts on Opsgenie only when we can consider  tuned enough.

infantesimone (changes requested):

src/domains/ecommerce-common/00_alerts.tf
pietro-tota: fixed with https://github.com/pagopa/pagopa-infra/pull/1956/commits/fb0a8d17193737295f100d5c7b0c33bf396b5922

infantesimone (dismissed):

infantesimone (approved):

PR #1965, opened 2024-04-23
Fix liquibase properties on value date

marimpietri-acn:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica (commented):
@aomegax is it right change changset and not create new one ?

marimpietri-acn:
the changeset is to be replaced because we need to choose the right property for the dbms(oracle, postgres, h2) that we want to update

PR #1993, opened 2024-05-07
[CHK-2822]: fix(ecommerce-for-io): delete transaction jwt validation

ciuffagianluca:
Add checking fro userId in jwt token, both for ecommerce-for-io api and for webview api.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-io/v1/_auth_request.xml.tpl
```diff
@@ -2,7 +2,6 @@
     <inbound>
         <base />
         <set-header name="x-pgs-id" exists-action="delete" />
-        <set-header name="x-user-id" exists-action="delete" />
```
infantesimone, line 5: Why is it deleted? this header must be set by the policy from sessionToken (only for ecommerce flow with jwt), and then passing it to the underlying service

ciuffagianluca: This field is set in the base policy. If you remove the header here it will result in 404 since now x-user-id is used to check the user of the transactions.

infantesimone (approved):

infantesimone (dismissed):

infantesimone (approved):

infantesimone (approved):

PR #2025, opened 2024-05-15
chore: LstTrx pdv pre-req

pasqualespica:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pasqualespica:
no longer necessary, transaction list users use CF clear in the requests and it is not necessary using PDV since the information is not shared with Data Lake

PR #2125, opened 2024-06-11
feat(dismissione-wisp): Add nodo-servicebus [NOD-944] -> [PAGOPA-1835]

lorenzo-catalano:
### List of changes
- `pagopa-{env}-weu-nodo-servicebus` (aka _**wispconverter**_ servicebus) creation
- `paainviart` retry and `Payment-Timeout` queues creation
- secret for `pagopa-{env}-weu-nodo-servicebus` (aka _**wispconverter**_ servicebus) creation
- Variable configuration (partitioning: https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning -> Not using a partition key -> round-robin fashion)

wispconverter receipts cosmos container ❌ 

### Service bus queue arguments 👀 
`azurerm_servicebus_queue` additional arguments discussion:
- `enable_partitioning` ✅
- `auto_delete_on_idle` ❌  out-of-scope: refers to the queues created on the fly, it doesn't fit into the use cases we would like to use it for [[ref1](https://learn.microsoft.com/en-us/azure/service-bus-messaging/advanced-features-overview#autodelete-on-idle), [ref2](https://stackoverflow.com/questions/76300209/azure-service-bus-auto-delete-on-idle-use-cases)]
- `default_message_ttl` ❌  the [default](https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-expiration#entity-level-expiration) is already appropriate for each intended use case 
> The default time-to-live value for a brokered message is the largest possible value for a signed 64-bit integer if not otherwise specified.

### Apply 🦾 
```
sh terraform.sh apply weu-dev --target=azurerm_servicebus_queue.wisp_converter_payment_timeout --target=azurerm_servicebus_queue_authorization_rule.wisp_converter_payment_timeout_consumer
```

`next-core` apply includes `azurerm_servicebus_queue_authorization_rule`, `azurerm_servicebus_queue`, `azurerm_servicebus_namespace` resources on `05_service_bus.tf`
```
sh terraform.sh apply dev --target=azurerm_servicebus_queue_authorization_rule.service_bus_01_queue_authorization_rule
```

### Test 🧪 
- init 1: create { `pagopa-d-weu-core-servicebus-01`, `nodo_wisp_paainviart_queue`, `wisp_converter_paainviart` } ✅
- increment: create/add { `nodo_wisp_payment_timeout_queue`, `wisp_converter_payment_timeout` } ✅
- update `manage` properties { `wisp_converter_payment_timeout`, `wisp_converter_paainviart` } ✅

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/nodo-common/06_service_bus_wisp_converter.tf
```diff
@@ -0,0 +1,70 @@
...
+  ]
+}
+
+resource "azurerm_servicebus_namespace_authorization_rule" "wisp_converter_servicebus_ns_auth" {
+  count = var.enable_wisp_converter ? 1 : 0
+
+  name         = "${local.project}-servicebus-ns-manager"
+  namespace_id = azurerm_servicebus_namespace.wisp_converter_servicebus[0].id
+
+  listen = true
```
aomegax, line 22: perhaps, like event-hub, it would be better to decouple reading and writing

AngeloCaporaso: It was added an advice as rule to follow

pasqualespica (approved):
jesc 🚀

PR #2205, opened 2024-06-25
fix: Set `additionalPaymentInformations` timestamps to naive date time format

giovanniberti:
### List of changes

Changed `closePayment` OpenAPI to use local datetime format in `additionalPaymentInformation` timestamps (`timestampOperation` field)

### Motivation and context

This PR fixes a regression on the `closePayment` call, previously the format specified did break the following calls.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/core/api/nodopagamenti_api/nodoPerPM/v2/_openapi.json.tpl
+++ b/src/core/api/nodopagamenti_api/nodoPerPM/v2/_openapi.json.tpl
@@ -207,7 +207,7 @@
           },
           "timestampOperation": {
             "type": "string",
-            "format": "date-time",
+            "format": "yyyy-MM-ddThh:mm:ss",
             "description": "Transaction timestamp"
           },
           "authorizationCode": {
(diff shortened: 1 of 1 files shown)


pietro-tota (approved):

src/core/alert_node_forwarder_apimv2.tf
```diff
@@ -1,7 +1,7 @@
 
 
 resource "azurerm_monitor_scheduled_query_rules_alert" "opex_pagopa-node-forwarder-responsetime-upd-v2" {
-  count               = var.env_short == "p" ? 1 : 0
+  count               = (var.env_short == "p" && var.enabled_features.apim_v2) ? 1 : 0
```
infantesimone, line 4: why in this PR?

giovanniberti: Moving to separate PR

infantesimone (approved):

src/core/alert_node_forwarder_apimv2.tf
giovanniberti: Moved to #2210

PR #2262, opened 2024-07-08
[CHK-3028] fix(session-wallet): return 401 if PM responds with status code different from 200

giovanniberti:
### List of changes

* ~check if session is created with PM before early returning from PM errors~
* return 401 if PM responds with status code different from 200

### Motivation and context

This PR fixes a bug where an expired wallet token created for the new authentication flow would make the API endpoint return 502 instead of 401.
This happens because PM returns 500 in case of an expired wallet token rather than 401, ~to sidestep this we don't return an early response in case of sessions with the new auth scheme.~, (we can't sidestep because PM gives us the user's fiscal code) we remap all status code that are not 200 into 401.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


ciuffagianluca (dismissed):

ciuffagianluca (approved):
🚀 well done!

giovanniberti:
Closed in favour of pagopa/pagopa-ecommerce-transactions-service#540

giovanniberti:
Closed in favour of https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/540

PR #2269, opened 2024-07-09
refactor(ecommerce-for-io): Use eCommerce-provided session token in transaction endpoints

giovanniberti:
### List of changes

 * allow eCommerce transaction endpoints to use the eCommerce-provided session token

### Motivation and context

This PR is a transitional work-in-progress to allow IO app to migrate to a flow where the shared platform token is used only to initialize the transaction.
Right now the platform token is used throughout the whole process, and we are targeting the same behaviour of the Checkout API, i.e. the transaction initialization (`POST /transactions`) generates a token that shall be used for all following API calls in the context of the same transaction.
With this PR the caller can use either the eCommerce-provided token or the platform token interchangeably (this is done for backward compatibility), in the future we will remove this dual authorization and allow only the eCommerce-provided token to be used.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


giovanniberti:
Closed in favour of pagopa/pagopa-ecommerce-transactions-service#540

PR #2305, opened 2024-07-17
chore: CHK-2997 delete old (unused) api for ecommerce and wallet

infantesimone:
### Motivation and context

The goal of this PR is to delete api version group related to old version that will not use in production

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


infantesimone:
task replaced by https://github.com/pagopa/pagopa-infra/pull/2384

PR #2353, opened 2024-07-31
[PAGOPA-2001] fix: resolved bug on KO sendpaymentresult on WISP dismantling

andrea-deri:
This PR contains a little fix for a wrong access on WISP dismantling's outbound policy fragment on `sendPaymentResultV2`. In particular, the extraction of the payment tokens from the request was wrongly executed because the nested object were different from `closePaymentV2` request, the one where the policy was "copied". The fix now allows a correct extraction of nested objects and a correct evaluation of the payment tokens.

#### Apply
The changes were applied using the following command:  
```
sh terraform.sh apply ENV \
  --target=azurerm_api_management_api_operation_policy.send_payment_result_api_v2_wisp_policy
```

### List of changes
 - Resolved bug on `sendPaymentResultV2` policy for payment token extraction

### Motivation and context
This fix is required in order to resolve a bug on WISP dismantling process.

### Type of changes
- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
🚀

PR #2360, opened 2024-08-01
feat: add x client id from client header

pietro-tota:
### List of changes
- Add `x-client-id-from-client` header handling on eCommerce Checkout POST transactions in order to trace EC cart vs Wisp redirect transactions
- Add clientId value into POST carts policies for discriminate WISP redirect from EC carts api calls
- Propagate client id discrimination to Checkout f.e. redirection

### Motivation and context
Those modification is needed in order to properly allow tracing between CHECKOUT_CART vs WISP_REDIRECT transactions

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


giovanniberti (dismissed):

petretiandrea (dismissed):

src/domains/checkout-app/api/checkout/checkout_carts_auth/v1/_base_policy.xml.tpl
```diff
@@ -4,6 +4,9 @@
       <base />
       <set-variable name="blueDeploymentPrefix" value="@(context.Request.Headers.GetValueOrDefault("deployment","").Contains("blue")?"/beta":"")" />
       <set-backend-service base-url="@("https://${ecommerce_ingress_hostname}"+context.Variables["blueDeploymentPrefix"]+"/pagopa-ecommerce-payment-requests-service")"/>
+      <set-header name="x-client-id" exists-action="override">
```
infantesimone, line 7: pay attention, this api is not dedicated exclusively to `WISP_REDIRECT`, for now it is the only client but soon we could give this api to other clients (like SEND). So you need to introduce a logic to differentiate exactly who the client is, adding a link to the user / API key (or another logic, you decide/propose :-) )

pietro-tota: You're absolutely right. I was thinking about posticipate it since SEND integration, but maybe is better to address it now! working on it 😄

pietro-tota: Addressed [here](https://github.com/pagopa/pagopa-infra/pull/2360/commits/442952fc2c97297a06b1a8d12a36df679c88b5d8).

Group names can be modified as needed.

For SEND flows maybe we can use another internal client id (that will be remapped to CHECKOUT_CART for Nodo/GEC api calls as will be done for WISP_REDIRECT) that will allow us to perform finer check against update status dashboard and discriminate SEND from WISP_REDIRECT carts initiated transactions :D

infantesimone: my comment comes from the idea to add a check who integrates with this api :-) `send` is just an example, let's remove it for now. Have you considered using `context.Subscription.Id` rather than creating groups? groups are fine, just a challenge to understand if it is useful to provide a group or not

pietro-tota: @infantesimone something like [that](https://github.com/pagopa/pagopa-infra/pull/2360/commits/78bd04c3fe3941364c15b6eb310dcb4fbd6ea0b4)?  I tested it successfully in DEV env. I put here a contains on the subscription id rather than an equality check since I checked already created subscriptions for every environments and they have a different id name but all contains `wisp` characters

infantesimone: Yes, we could consider something like that, but here we need @pasqualespica @aomegax  support, how are subscriptions  configured in Prod for carts/wisp? What is the subscription.id?

aomegax: I think it's better to unify the `subscription.id` value to `wisp-dismantling-checkout`.

Or, on the other hand, set the value on the client bypassing the APIM policy.

pasqualespica: @infantesimone for`What is the subscription.id?` see  [here](https://learn.microsoft.com/en-us/powershell/module/az.accounts/set-azcontext?view=azps-12.1.0#-subscription)

tnx @pietro-tota @aomegax i thins its good idea

pietro-tota: Thanks @aomegax @pasqualespica.

@aomegax I would avoid pass `x-client-id` from the client:

the main aim here is to _derive_ the `POST carts` client based on the subscription key used for the api call, allowing set this header from an external client could expose us to one client set the header to `WISP_REDIRECT` even if it's not the wisp dismantling client itself. Even if this would be a bug we cannot perform any coherence check against an externally received value for that header, using subscription ID matching, instead, is client transparent and a 1-to-1 match :) 
What do you think?

There could be a mismatch also with this configuration if the same subscription key is used by different clients but I think that we exclude this case: any client MUST use it's subscription key that will have different id 

@infantesimone add `subscrption.id` exact match [here](https://github.com/pagopa/pagopa-infra/pull/2360/commits/991244a6bbd9369d016b622aa0dc71a2d5a10f3c) based on the above comment

infantesimone: > Thanks @aomegax @pasqualespica.
> 
> @aomegax I would avoid pass `x-client-id` from the client:
> 
> the main aim here is to _derive_ the `POST carts` client based on the subscription key used for the api call, allowing set this header from an external client could expose us to one client set the header to `WISP_REDIRECT` even if it's not the wisp dismantling client itself. Even if this would be a bug/misconfiguration we cannot perform any coherence check against an externally received value for that header, using subscription ID matching, instead, is client transparent and a 1-to-1 match :) What do you think?
> 
> There could be a mismatch also with this configuration if the same subscription key is used by different clients but I think that we exclude this case: any client MUST use it's subscription key that will have different id
> 
> @infantesimone add `subscrption.id` exact match [here](https://github.com/pagopa/pagopa-infra/pull/2360/commits/991244a6bbd9369d016b622aa0dc71a2d5a10f3c) based on the above comments

sounds good! but in PROD do we have to update the subscription key cc @aomegax @pasqualespica? because it is associated to a nominal user, it could be the opportunity to associate it to a team or technical user.

src/domains/ecommerce-app/api/ecommerce-checkout/v2/_openapi.json.tpl
```diff
@@ -57,6 +57,20 @@
               "format": "uuid"
             }
           },
+          {
+            "in": "header",
+            "name": "x-client-id-from-client",
```
infantesimone, line 62: This header should not be exposed but is internal to the service

infantesimone (changes requested):

src/domains/ecommerce-app/api/ecommerce-checkout/v2/_openapi.json.tpl
pietro-tota: this header is valued using the client id query parameter injected by payment request and then passed back from f.e. into POST transaction api call so that b.e. can save this information.

This have been done since in POST transactions coming from a cart no information about original cart id is passed to transactions-service so the other way can be to add carts reading capability to transactions service that can retrieve the client id associated to that cart

infantesimone (approved):

PR #2390, opened 2024-08-28
chore: remove keys used by PGS for eCommerce

giovanniberti:
### List of changes

 * remove keys used by PGS for eCommerce
 * remove CDN rules associated to PGS

### Motivation and context

These keys were used only by transactions with XPAY and VPOS gateways which are no longer used.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---

Depends on pagopa/pagopa-ecommerce-event-dispatcher-service#323
Depends on pagopa/pagopa-ecommerce-transactions-service#561
Depends on #2401


infantesimone (changes requested):
Wait :-) this API is used in prod

infantesimone:
there is a mismatch between the PR and the description :-)

giovanniberti:
Updated 👌

src/domains/ecommerce-common/02_security.tf
```diff
@@ -257,18 +257,6 @@ resource "azurerm_key_vault_secret" "notifications_service_api_key" {
   }
 }
 
-resource "azurerm_key_vault_secret" "payment_transactions_gateway_api_key" {
```
infantesimone, line 260: this api key is used https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/aa0eabb4d0107fa92cd027cde1d6fa18e212ed6f/helm/values-prod.yaml#L191.
so this PR can be applied after the `transaction-service` no longer uses this api key, right?

giovanniberti: Yes, I thought I had already added the dependencies but I didn't, udpated them in the description right now :)

petretiandrea (approved):

infantesimone (approved):

infantesimone (approved):

PR #2401, opened 2024-09-03
refactor: round 2 removal of PGS APIs

giovanniberti:
### Motivation and context

This PR removes the remaining PGS resources inside the `core` domain

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


petretiandrea (approved):

src/core/apim_payment_transactions_gateway.tf
```diff
@@ -1,249 +0,0 @@
...
-    # params for all api versions
-    display_name          = "Payment Transactions Gateway - payment-transactions-gateway"
-    description           = "RESTful APIs provided to support payments with payment gateways"
-    path                  = "payment-transactions-gateway/internal"
-    subscription_required = true
-    service_url           = null
-  }
-}
-
-resource "azurerm_api_management_api_version_set" "payment_transactions_gateway_internal_api" {
```
infantesimone, line 104: this resource expose API for bancomatpay flows invoked by the IO app, have you verified no traffic on this api?

infantesimone (changes requested):

src/core/apim_payment_transactions_gateway.tf
giovanniberti: Given that there is a planned forced update on the IO app and this API specifically is currently getting less than 100 requests/day (gathered from APIM data pertaining last week), so I assume we are safe to remove this.
If we're not I'll remove it from this PR

infantesimone (approved):

PR #2402, opened 2024-09-05
[PPANTT-106] feat: Introduced liquidbase changelog for data_creazione…

alessio-cialini:
… on station model

### List of changes

- Introducing data_creazione property to stazioni withing cfg database

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


jacopocarlini (changes requested):
you need to create the 3.31.0 in all schema.
Create a empty 3.31.0 version in all directories inside the `changelog` folder.
[here](https://github.com/pagopa/pagopa-infra/blob/main/src/psql/nodo/liquibase/changelog/online/db.changelog-master-3.8.0.xml) an example

PR #2404, opened 2024-09-06
[CHK-3213] feat: add new wallet log events collection

CianoDanilo:
### List of changes
Added a new collection which contains logging events of payment wallet service

### Motivation and context

The new logging events collection will contain the new and more detailed logging event of payment wallet service, making previous logging events collection deprecated and eligible for a future removal
### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/pay-wallet-common/03_cosmosdb.tf
+++ b/src/domains/pay-wallet-common/03_cosmosdb.tf
@@ -136,20 +136,6 @@ locals {
       ]
       shard_key = "walletId"
     },
-    { # collection with detailed events
-      name = "payment-wallets-log-events"
-      default_ttl_seconds = "600"
-      indexes = [{
-        keys   = ["_id"]
-        unique = true
-        },
-        {
-          keys   = ["walletId", "timestamp"]
-          unique = true
-        }
-      ]
-      shard_key = "walletId"
-    },
   ]
 }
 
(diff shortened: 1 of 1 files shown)


src/domains/pay-wallet-common/03_cosmosdb.tf
```diff
@@ -155,6 +155,30 @@ module "cosmosdb_pay_wallet_collections" {
   lock_enable = var.env_short != "p" ? false : true
 }
 
+module "cosmosdb_payment_wallets_log_events_collection" {
```
pietro-tota, line 158: do not duplicate this module, use the above `cosmosdb_pay_wallet_collections` one.
You can add a `default_ttl_seconds` parameter for each locals collection parameter valued with null (the [default](https://github.com/pagopa/terraform-azurerm-v3/blob/7dbbc06d591d3ce66536a7bdb2208b1370de04dd/cosmosdb_mongodb_collection/variables.tf#L30-L34) value for the module) and set the 2592000 value for this new collection

CianoDanilo: Together with @infantesimone we valued more important not modifying already existing collections since terraform would consider adding the default ttl an update; we prefer to not have any impact on already existing collections in prod env

pietro-tota: IDK, I just made the modification locally adding parameter and it gives only diffs in one collection (the one where i set the attribute with a value different from null for test purpose)

pietro-tota: Considering the module default value is null we can set it explicitly to null for the already existent collections and value it only for the new one. This way the stored terraform state for the already existent collection will remain the same and can configure TTL only where needed (for future collections or when we will set it for the already existent ones 😄)

pietro-tota (approved):

src/domains/pay-wallet-common/03_cosmosdb.tf
CianoDanilo: Proposal applied in 8fe27322

pietro-tota (dismissed):

petretiandrea (dismissed):

src/domains/pay-wallet-common/03_cosmosdb.tf
```diff
@@ -136,6 +143,20 @@ locals {
       ]
       shard_key = "walletId"
     },
+    { # collection with new and detailed logging events
+      name                = "payment-wallets-log-events"
+      default_ttl_seconds = "2592000" #30 days
```
infantesimone, line 148: why 30 days? if we estimate how many events we can have per month maybe we can extend to 90 days. So larger time window for support.

infantesimone (changes requested):

src/domains/pay-wallet-common/03_cosmosdb.tf
CianoDanilo: Considering that the current events collection, that has been created about 2 weeks ago, contains about 560k documents, we could estimate roughly 1.2M events per month; could we extend the TTL to 90 days given this estimation?

infantesimone: I propose at least 60 days which is enough time to recover any events for debugging,
"ortofruttifero" poll? what you think? @pagopa/pagopa-team-touchpoint

pietro-tota: 🍎

infantesimone (approved):

PR #2425, opened 2024-09-18
feat: [PAYMCLOUD-35] add infra opsgenie action group

ffppa:
### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/next-core/00_key_vault.tf
```diff
@@ -164,3 +164,11 @@ data "azurerm_key_vault_secret" "sec_storage_id" {
   name         = "sec-storage-id"
   key_vault_id = module.key_vault.id
 }
+
+# INFRA OpsGenie PagoPA_Azure_infra webhook key
+data "azurerm_key_vault_secret" "opsgenie_infra_webhook_key" {
+  count = var.env_short == "p" ? 1 : 0
+  name  = "opsgenie-infra-webhook-token"
```
pasqualespica, line 171: where did you create the secret?

pasqualespica (dismissed):

src/next-core/00_key_vault.tf
ffppa: I requested that the secret be created in the prod keyvault in order to because the action group will only have to be entered there for the finder service, the webhook key has already been entered

mamari90 (approved):

PR #2426, opened 2024-09-18
feat: [CHK-3244] Update open api v2 to include an optional "lang" header

EmanueleBVtech:
### List of changes

- Updated open api V1 for ecommerce checkout to include an optional header for language preference

### Motivation and context
This change is part of a bigger task aiming to provide NPG with the language preference as set by the user on the checkout FE.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-checkout/v2/_openapi.json.tpl
```diff
@@ -28,6 +28,77 @@
     "description": "Design review"
   },
   "paths": {
+    "/payment-methods/{id}/sessions": {
```
pietro-tota, line 31: why add this v2 api version? an optional header have been added to api so no breaking changes.
optional header can be add to the v1 version.

what do you think?

EmanueleBVtech: apply the new header only to v2 (and eventually move from v1 to v2 the required APIs) was decided during the refinement of the task.

src/domains/ecommerce-app/api/ecommerce-checkout/v2/_openapi.json.tpl
```diff
@@ -28,6 +28,77 @@
...
+            "description": "Recaptcha response",
+            "schema": {
+              "type": "string"
+            },
+            "required": true
+          },
+		  {
+            "in": "header",
+            "name": "lang",
+            "required": true,
```
giovanniberti, line 61: Shouldn't this be optional?

EmanueleBVtech: Yes as you can see from previous commits it was initially optional,
but i had to set it as required due to an issue with "yarn generate" on the checkout repo wich caused the optional language header to be translated into code with a bug in it (it would not compile due to sintax).

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_openapi.json.tpl
```diff
@@ -661,6 +661,15 @@
...
           },
+          {
+            "in": "header",
+            "name": "lang",
+            "required": true,
+            "description": "Language requested by the user",
+            "schema": {
+              "type": "string",
+            }
+          },
```
pietro-tota, line 672: the `lang` header should be add also for POST /transactions/{transactionId}/auth-requests api call or POST /transactions one to propagate user language to transactions-service b.e.?

POST sessions api call is performed for guest cards payment only so this information have to be propagated also for transactions performed with other methods than cards

pietro-tota (approved):
lgmt 🚀

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_openapi.json.tpl
```diff
@@ -389,7 +389,16 @@
             },
             "required": true,
             "description": "Transaction ID"
-          }
+          },
+          {
+            "in": "header",
+            "name": "lang",
+            "required": true,
```
infantesimone, line 396: why required?

EmanueleBVtech: initially this was not required on FE side, but at compile time we get a sintax error after using 'yarn generate'.

the generated code contains an error that blocks the compiling of the app, 
if we use required: true, the error does not happen.

infantesimone: maybe without             
```
"required": true,
```

it's attempt to "raggirare" codegen :)

EmanueleBVtech: i will try this on a dummy branch to see if it helps

EmanueleBVtech: no luck, same result
image

EmanueleBVtech: this + updating the library on FE did the trick, thanks for the hint

EmanueleBVtech: didn't work, even after updating the codegen lib :(

pietro-tota (approved):

PR #2428, opened 2024-09-18
fix: [PAGOPA-2155] Added check on nodoInviaRPT/nodoInviaCarrelloRPT outbound policy

aomegax:
### List of changes
- `wisp_nodoInviaRPT_nodoInviaCarrelloRPT_outbound_policy.xml` fragment

### Motivation and context
When a primitive arrives, the wisp_sessionId will is set and the rpt/timer POST is executed even if request does not belong to WISP dismantling.
So the check avoid to execute unuseful steps.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


andrea-deri (approved):
🔝

PR #2433, opened 2024-09-20
[PAGOPA-2158] fix: add session id for better debug in D-Wisp alerts

andrea-deri:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


andrea-deri:
Closed because it was never completed.

PR #2435, opened 2024-09-23
feat(retry-policy): [CHK-3228] add retry on backend failure to npg notification policies

CianoDanilo:
### List of changes
Added a retry behaviour to APIM policies for npg notification apis. The retry will be triggered on backend failures and only for certain type of errors.

### Motivation and context

We want to provide with a retry mechanism those apis which aren't covered by a retry logic by the client and for those errors that are retryable (i.e. a connection in the pool has been terminated due to inactivity resulting in a 500 status code)
### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


ciuffagianluca (dismissed):

src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl
```diff
@@ -159,7 +159,10 @@
         <!-- end send transactions service PATCH request -->
     </inbound>
     <backend>
-        <forward-request timeout="10" />
+        <retry condition="@(context.Response.StatusCode >= 500)"
+              interval="1" count="2" first-fast-retry="true">
```
pietro-tota, line 163:
```suggestion
              interval="2" count="3" first-fast-retry="true">
```

What about use the same configuration used for [transactions-service retry policy](https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/9a895f367227770c53f403da0c6f09490a1b33a5/src/main/resources/application.yml#L4-L5) here too?

pietro-tota (approved):
Well done, just a consideration :)

src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl
CianoDanilo: We could do it, do we know what it the client's timeout?

CianoDanilo: Applied in [f4c1f92]

pietro-tota (approved):

PR #2438, opened 2024-09-24
fix: [PAYMCLOUD-92] Fix source and modified path filter excluding scripts folder

ffppa:
### List of changes

This changes exclude scripts dir from terraform checks to avoid failure

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


mamari90 (dismissed):
👍

mamari90 (approved):
🥇

PR #2444, opened 2024-09-26
fix: 3.31.0 updated for cfg-partner with missing fields

alessio-cialini:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/psql/nodo/liquibase/changelog/cfg/db.changelog-master-3.31.0.xml
+++ b/src/psql/nodo/liquibase/changelog/cfg/db.changelog-master-3.31.0.xml
@@ -7,6 +7,9 @@
 
   <include file="./db.changelog-master-3.30.0.xml"/>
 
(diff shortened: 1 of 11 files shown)


src/psql/nodo/liquibase/changelog/cfg/db.changelog-master-3.31.0.xml
```diff
@@ -7,6 +7,9 @@
 
   <include file="./db.changelog-master-3.30.0.xml"/>
 
+  <include file="./${version}/db.changelog-20240711000000_add_field_aca.xml" labels="${version}"/>
```
pasqualespica, line 10: @alessio-cialini this is not need for `cfg` its already included. into `db.changelog-master-3.30.0.xml`

pasqualespica (changes requested):

pasqualespica (approved):

PR #2446, opened 2024-09-27
feat: (alerts) CHK-3231 Modify the query to make availability traffic-dependent

ciuffagianluca:
In order to reduce false alarm fired by alerts, in this PR we try to change availability according to the traffic 

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -48,14 +48,23 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_for_checkout_a
   description    = "eCommerce Availability less than or equal 99%"
   enabled        = true
   query = (<<-QUERY
+let thresholdTrafficMin = 40;
+let thresholdTrafficLinear = 100;
+let lowTrafficAvailability = 80;
```
pietro-tota, line 53:
```suggestion
let lowTrafficAvailability = 90;
```
what about set this threshold to the one actually used for alert?

pietro-tota: I would set this lowTrafficAvaliability to 90 also for all other alerts as for now, and perform tuning if needed in future, what do you think?

pietro-tota (dismissed):
LGTM 🚀  just a consideration about min availability threshold

pietro-tota (dismissed):

pietro-tota (approved):

PR #2457, opened 2024-10-02
feat(ecommerce user stats): [CHK-3293] add PUT user stats api V1

pietro-tota:
### List of changes
Add api for eCommerce user-stats service.
Those api includes a PUT api call that will be used to update user last used method during an eCommerce transaction

### Motivation and context
This api will be used by eCommerce event-dispatcher to signal the latest used method for an user

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


ciuffagianluca (approved):

src/domains/ecommerce-app/api/ecommerce-user-stats-service/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,228 @@
...
+      "post": {
+        "parameters": [
+          {
+            "name": "userId",
+            "in": "path",
+            "required": true,
+            "description": "User identifier",
+            "schema": {
+              "type": "string",
+              "format": "uuid"
```
ciuffagianluca, line 46:
```suggestion
```
Remove this format declaration, since userId is not a UUID

src/domains/ecommerce-app/api/ecommerce-user-stats-service/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,228 @@
...
+      "ApiKeyAuth": []
+    }
+  ],
+  "paths": {
+    "/user/lastPaymentMethodUsed": {
+      "post": {
+        "parameters": [
+          {
+            "in": "header",
+            "name": "x-user-id",
```
infantesimone, line 41: curiosity, why in the header and not in the body?

infantesimone: in the `header` it is useful when the data is used at apim level, in this case is there an authorization mechanism that identifies that api can be called for that payment flow? so for a given transaction?

pietro-tota: This POST api call will be called by event-dispatcher in a b2b interaction, the only identified auth mechanism here is key.

The GET api call, instead, will be called by IO app (and so authenticated with session token, so user id will be injected to the service as header parameter).

I put this information here to have _homogeneity_ between the POST and GET api calls, but it can be moved inside POST body, the modification is pretty simple here.

What do you think?

PR #2465, opened 2024-10-04
feat: [CHK-3298] add user stats api key in keyvault

ciuffagianluca:
Add key-vault entry to store user-stats api key

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pietro-tota (dismissed):

src/domains/ecommerce-common/02_security.tf
```diff
@@ -486,6 +486,19 @@ resource "azurerm_key_vault_secret" "transactions_service_auth_update_api_key" {
   }
 }
 
+
+resource "azurerm_key_vault_secret" "user_stats_api_key" {
+  name         = "user-stats-api-key"
```
infantesimone, line 491: user-stats is an ecommerce microservice, who will invoke this api? I recommend adding a reference to the client in the name, the goal is to avoid using this api key unintentionally. If there are multiple clients internal or external to ecommerce, it is always better to differentiate the api keys

infantesimone (approved):

PR #2490, opened 2024-10-16
fix: sync call to rpt timer WISP Conv PAGOPA-2292

FedericoRuzzier:
### List of changes
Changing call create rpt timer in wisp conv from async to sync
`sh terraform.sh apply <env> -target=azapi_resource.wisp_nodoinviarpt_nodoinviacarrellorpt_outbound_policy -target=terraform_data.sha256_nodoinviarpt_wisp_nodoinviacarrellorpt_outbound_policy`
Update alert, exclude wic error 1300
`sh terraform.sh apply weu-prod -target=azurerm_monitor_scheduled_query_rules_alert.opex_pagopa-wisp-converter-wic-error`

### Motivation and context

Needed in order to correctly execute flow.

The request cannot be executed in one way mode because timer creation could be consecutive to the timer deletion in the case of unfortunate interleaving in wisp-converter.

There cannot be concurrency between the `POST /rpt/timer` request and the `GET /wisp-converter/redirect/api/v1/payments` request (these are 2 calls that the user might activate closely, <1s).

### Type of changes

- [ ] Add new resources

This change could add a delay to the call which depends on the synchronism of `/rpt/timer` request

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/core/api/nodopagamenti_api/nodoPerPa/v1/wisp_nodoInviaRPT_nodoInviaCarrelloRPT_outbound_policy.xml
+++ b/src/core/api/nodopagamenti_api/nodoPerPa/v1/wisp_nodoInviaRPT_nodoInviaCarrelloRPT_outbound_policy.xml
@@ -24,7 +24,7 @@
       <when condition="@(!((string)context.Variables["wisp_sessionId"]).Equals("NONE"))">
       <cache-store-value key="@("wisp_timer_hang_" + (string)context.Variables["wisp_sessionId"])" value="@(DateTime.UtcNow.AddSeconds({{wisp-checkout-predefined-expiration-time}}).ToString())" duration="{{wisp-checkout-predefined-expiration-time}}" caching-type="external" />
       <!-- set rpt/timer to check if redirect will be executed successfully -->
-      <send-request mode="new" response-variable-name="rpt-timer-create-response">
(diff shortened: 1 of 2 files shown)


src/core/api/nodopagamenti_api/nodoPerPa/v1/wisp_nodoInviaRPT_nodoInviaCarrelloRPT_outbound_policy.xml
```diff
@@ -24,7 +24,7 @@
       <when condition="@(!((string)context.Variables["wisp_sessionId"]).Equals("NONE"))">
       <cache-store-value key="@("wisp_timer_hang_" + (string)context.Variables["wisp_sessionId"])" value="@(DateTime.UtcNow.AddSeconds({{wisp-checkout-predefined-expiration-time}}).ToString())" duration="{{wisp-checkout-predefined-expiration-time}}" caching-type="external" />
       <!-- set rpt/timer to check if redirect will be executed successfully -->
-      <send-one-way-request mode="new">
+      <send-request mode="new" response-variable-name="rpt-timer-create-response">
```
aomegax, line 27: set `ignore-error="true"`

pasqualespica (approved):

aomegax (approved):

PR #2492, opened 2024-10-17
feat: [CHK-3341] add user stats availability alert

scaminati-bv:
### List of changes
- add availability alert for user-stats-service apis
- add custom_webhook_payload json into action with opsgenie alert values

### Motivation and context
The goal of this PR is to add availability alert to all user-stats-service apis

### Type of changes

- [X] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-common/00_alerts.tf
EmanueleBVtech, line 1: Instead of hardcoding values like thresholdTrafficMin, thresholdTrafficLinear, lowTrafficAvailability, and highTrafficAvailability, consider making them variables. This will make the script more flexible and easier to adjust without modifying the code.

EmanueleBVtech, line 1: Just a suggestion. Break down the query into smaller, well-commented sections to enhance readability and maintainability.

src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -254,4 +254,49 @@ AzureDiagnostics
...
+# eCommerce user stats service availability
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_user_stats_service_availability_alert" {
+  count = var.env_short == "p" ? 1 : 0
+
+  name                = "ecommerce-user-stats-service-availability-alert"
+  resource_group_name = azurerm_resource_group.rg_ecommerce_alerts[0].name
+  location            = var.location
+
+  action {
+    action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id]
```
EmanueleBVtech, line 268: Include more detailed information in the alert payload to help the team quickly understand the issue.

```
action {
  action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id]
  email_subject          = "[eCommerce] User stats service availability less than 99%"
  custom_webhook_payload = jsonencode({
    "alertName" : "ecommerce-user-stats-service-availability-alert",
    "severity" : "1",
    "description" : "eCommerce User stats service availability less than 98% in the last 30 minutes detected",
    "affectedService" : "User Stats Service",
    "environment" : var.env_short
  })
}

`

File: src/domains/ecommerce-common/00_alerts.tf
Comment (scaminati-bv) [reply src/domains/ecommerce-common/00_alerts.tf:1]: These threshold could change between alerts. So we can't use global variables with the same values.

Comment (scaminati-bv) [reply src/domains/ecommerce-common/00_alerts.tf:1]: The structure of this query is the same as in the previous alert, and it also seems quite clear to me. Therefore I preferred to keep the same form.

Comment (EmanueleBVtech) [reply src/domains/ecommerce-common/00_alerts.tf:1]: I see their value is initialized and never updated, how are they gonna change?

File: src/domains/ecommerce-common/00_alerts.tf (hunk 1 above)
Comment (scaminati-bv) [reply src/domains/ecommerce-common/00_alerts.tf:268]: In all other actions the custom_webhook_payload property is always populated with _“{}”_, so it is most likely not needed.

File: src/domains/ecommerce-common/00_alerts.tf
Comment (EmanueleBVtech) [reply src/domains/ecommerce-common/00_alerts.tf:1]: I dont think we should avoid applying a (readability) improvement just to remain similar to older code.
Improvements also require this type of refactor, i think.

File: src/domains/ecommerce-common/00_alerts.tf [hunk 2]
```diff
@@ -254,4 +254,49 @@ AzureDiagnostics
...
+  enabled        = true
+  query = (<<-QUERY
+let thresholdTrafficMin = 200;
+let thresholdTrafficLinear = 500;
+let lowTrafficAvailability = 94;
+let highTrafficAvailability = 98;
+let thresholdDelta = thresholdTrafficLinear - thresholdTrafficMin;
+let availabilityDelta = highTrafficAvailability - lowTrafficAvailability;
+AzureDiagnostics
+| where url_s startswith 'https://api.dev.platform.pagopa.it/ecommerce/user-stats-service/v1'
```
ciuffagianluca, line 283:
```suggestion
| where url_s startswith 'https://api.dev.platform.pagopa.it/ecommerce/user-stats-service/v1'
```
| where url_s startswith 'https://api.platform.pagopa.it/ecommerce/user-stats-service/v1'

ciuffagianluca, line 283:
```suggestion
| where url_s startswith 'https://api.platform.pagopa.it/ecommerce/user-stats-service/v1'
```

src/domains/ecommerce-common/00_alerts.tf
ciuffagianluca: it could be a good point to use global variables for these values, but we could run into the risk of having many variables (potentially 4 for each alert) and have more difficult to read the query. So if we have to name the variables with something like _high_traffic_availability_98_ and _high_traffic_availability_99_ are we sure that it is a real improvement of readability? Please, consider careful this point.

ciuffagianluca: It's a good idea to insert these comment lines into the query to improve the readability. It's really clear for us that work with this query by months, but it couldn't be the same for someone new for this project. My only doubt is about azure, if it runs query with comments inside or not. Please verify this point, and if yes, comments are welcome

src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -254,4 +254,49 @@ AzureDiagnostics
...
+| summarize
+    Total=count(),
+    Success=countif(responseCode_d < 500 and DurationMs < 500)
+    by Time = bin(TimeGenerated, 15m)
+| extend trafficUp = Total-thresholdTrafficMin
+| extend deltaRatio = todouble(todouble(trafficUp)/todouble(thresholdDelta))
+| extend expectedAvailability = iff(Total >= thresholdTrafficLinear, toreal(highTrafficAvailability), iff(Total <= thresholdTrafficMin, toreal(lowTrafficAvailability), (deltaRatio*(availabilityDelta))+lowTrafficAvailability)) 
+| extend Availability=((Success * 1.0) / Total) * 100
+| where Availability < expectedAvailability
```
ciuffagianluca, line 292: Maybe this query could be simpler than this. This kind of query is useful for those apis that is used in user experience traffic. The api of `lastIUsageUpdate` instead is used for intra service communication and could follow a _traditional_ query like this one:
`
AzureDiagnostics 
| where url_s matches regex "https://api.platform.pagopa.it/ecommerce/user-stats-service/v1"
| where method_s == "POST"
| where responseCode_d != 200 or DurationMs > 2000
| project TimeGenerated, responseCode_d, DurationMs
`

src/domains/ecommerce-common/00_alerts.tf
scaminati-bv: @ciuffagianluca Yes, it is possible to add comments on query ([DOC](https://learn.microsoft.com/en-us/kusto/query/comment?view=microsoft-fabric))

src/domains/ecommerce-common/00_alerts.tf
scaminati-bv: Ok, I used the same query as the one above because I imagined it in the same flow as the IO app.
But I agree with you that it is an intra service api so we can use the translational query

src/domains/ecommerce-common/00_alerts.tf
scaminati-bv: I will analyze the custom_webhook_payload  property and we could make a test with this value populated

scaminati-bv: added here 5544ec956736bec97bd63d4112ae48e0eed9c123

src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -254,4 +254,46 @@ AzureDiagnostics
...
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_user_stats_service_availability_alert" {
+  count = var.env_short == "p" ? 1 : 0
+
+  name                = "ecommerce-user-stats-service-availability-alert"
+  resource_group_name = azurerm_resource_group.rg_ecommerce_alerts[0].name
+  location            = var.location
+
+  action {
+    action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id]
+    email_subject          = "[eCommerce] User stats service availability less that 99%"
```
infantesimone, line 269: specify that it is not the availability of the `user-stats-service` but only of the PUT api

scaminati-bv: If we add more APIs to this group in the future, will we edit this description again?
Or we could modify the description to explain that they refer only to api reachable between microservices.

scaminati-bv: At the moment `where ` condition detect all user-stats-service api (only PUT is exposed.

`where url_s startswith 'https://api.platform.pagopa.it/ecommerce/user-stats-service/v1'`

infantesimone: but in this API group there is no `GET user-stats`, which is part of the service. If all GET fail but PUT continue to work, this alert does not fire but we can say that `[eCommerce] User stats service availability less than 99%` is ok?

infantesimone: anyway it's not blocking for me

scaminati-bv: done here 1ea2363a442065057f721759d9ad082b8adb20c7

PR #2540, opened 2024-11-12
feat: CHK-3498 remove recaptcha post sessions

ciuffagianluca:
Limit reCaptcha request removing it from api that doesn't expose sensitive data like POST `/sessions`

### List of changes

- remove recaptcha for POST `/sessions`
- rename policy file

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-app/api/ecommerce-checkout/v1/_post_sessions.xml.tpl
```diff
@@ -0,0 +1,19 @@
+<policies>
+    <inbound>
+    <rate-limit-by-key 
+    calls="150"
```
CianoDanilo, line 4: Isn't the threshold of 150 calls per 10 seconds too high?

ciuffagianluca: We set this threshold together with @infantesimone reasoning about 2 points:

1. this is the same rate limit set in the base policy for payment-request verify
2. thinking about one office with a proxy to go out in internet that expose same ip for all machines

Considering these 2 points we decide to set this value for this threshold

EmanueleBVtech (dismissed):

CianoDanilo (approved):

pietro-tota (approved):
🚀 well done!

src/domains/ecommerce-app/README.md
```diff
@@ -1,5 +1,5 @@
 <!-- markdownlint-disable -->
-<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+<!-- BEGIN_TF_DOCS -->
```
pietro-tota, line 2: maybe here we have to rollback those pre-commit changes?

ciuffagianluca: Why should we rollback?

PR #2603, opened 2024-11-25
feat(recaptcha): CHK-3588 add new named value for ecommerce recaptcha

infantesimone:
### List of changes

- add new named value to support ecommerce for checkout recpatcha

### Motivation and context

The goal of this PR is to add new secret and named value `ecommerce_for_checkout_google_recaptcha_secret` to support google recaptcha enterprise migration.
The idea is to add new named value and not update it , in order to simplify an eventual rollback.

### Type of changes

- [x] Add new resources

- [x ] No

- [ ] Yes
- [x] No

### Other information

---


infantesimone:
ping @pagopa/pagopa-team-touchpoint

infantesimone:
the plan fails because the secret present in common has not been applied yet. For the apply process, you first need to create the secret and then create the named value.

pietro-tota (approved):

ciuffagianluca (dismissed):

PR #2606, opened 2024-11-26
feat(recaptcha): CHK-3589 Add double recaptcha check for ecommerce enterprise

infantesimone:
### List of changes

- add double check for recaptcha enterprise

### Motivation and context

The goal of this PR is to create a  policy that allows to manage and orchestrate the reCAPTCHA validation for both account versions. The policy should invoke reCAPTCHA using both the old and the new secret keys, and the validity test will be considered successful if at least one of the two reCAPTCHA calls returns a valid result. This fallback logic helps reduce the risk of service disruptions, as the system will still accept the validation if either of the two reCAPTCHAs (the new one or the previous one) is functioning.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/ecommerce-app/api/ecommerce-checkout/v1/_payment_request_policy.xml.tpl
+++ b/src/domains/ecommerce-app/api/ecommerce-checkout/v1/_payment_request_policy.xml.tpl
@@ -2,6 +2,7 @@
   <inbound>
     <!-- Check google reCAPTCHA token validity START -->
     <set-variable name="recaptchaSecret" value="{{google-recaptcha-secret}}" />
+    <set-variable name="recaptchaEnterpriseSecret" value="{{ecommerce-for-checkout-google-recaptcha-secret}}" />
     <set-variable name="recaptchaToken" value="@(context.Request.OriginalUrl.Query.GetValueOrDefault("recaptchaResponse"))" />
     <choose>
         <when condition="@(context.Variables["recaptchaToken"] == null || context.Variables["recaptchaToken"] == "")">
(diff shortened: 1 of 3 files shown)


ciuffagianluca (approved):

CianoDanilo (approved):

scaminati-bv (approved):

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_payment_request_policy.xml.tpl
```diff
@@ -21,9 +22,24 @@
     <set-variable name="recaptcha-check-body" value="@(((IResponse)context.Variables["recaptcha-check"]).Body.As<JObject>())" />
     <choose>
       <when condition="@(((IResponse)context.Variables["recaptcha-check"]).StatusCode != 200 || ((bool) ((JObject) context.Variables["recaptcha-check-body"])["success"]) != true)">
-        <return-response>
-          <set-status code="401" reason="Unauthorized" />
-          </return-response>
+          <!-- Double Check google reCAPTCHA Enterprise token validity START -->
+           <send-request ignore-error="true" timeout="10" response-variable-name="recaptcha-enterprise-check" mode="new">
```
pietro-tota, line 26: I would rather switch this "enterprise verification api call check" with the above one since once switch is performed is expected that all verifications will succeed with the enterprise (new) version and api call to the old one will be performed only for those api call coming from an un-updated checkout version (cached version maybe).
This way we will avoid a double api call for all the updated frontends.
What do you think?

pietro-tota (approved):
LGMT just an observation about who have to been called as fallback strategy

src/domains/ecommerce-app/api/ecommerce-checkout/v1/_payment_request_policy.xml.tpl
infantesimone: I thought about it, but this would mean applying the policy very close to the actual release. Otherwise, we would slightly affect the metrics on the new reCAPTCHA. So, we pay for an extra API call, but we don't impact the metrics on the new reCAPTCHA.

infantesimone: I might not have been very clear :-) The idea is to apply it in production a few hours before switching the frontend... this way, we can be more confident when doing the switch with the frontend. I wouldn't want traffic to hit the new reCAPTCHA with tokens from the old reCAPTCHA, to avoid affecting the metrics of the new reCAPTCHA.

PR #2608, opened 2024-11-26
feat(alerts): configure alerts for payment wallet ingestion

pietro-tota:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/observability/env/prod/terraform.tfvars
+++ b/src/domains/observability/env/prod/terraform.tfvars
@@ -259,13 +259,31 @@ ehns_metric_alerts = {
       }
     ],
   },
+  no_wallet_ingestion_alert = {
+    aggregation = "Total"
+    metric_name = "IncomingMessages"
+    description = "Payment wallet onboarding written events less than 2000 detected in the last 24h"
+    operator    = "LessThanOrEqual"
+    threshold   = 2000
+    frequency   = "PT1H"
+    window_size = "P1D"
+    dimension = [
+      {
+        name     = "EntityName"
+        operator = "Include"
+        values = [
+          "payment-wallet-evt-tx",
+        ]
+      }
+    ],
+  },
 }
 
 
 
 
 eventhubs_gpd = [
-    {
+  {
     name              = "test-evh" # test
     partitions        = 1
     message_retention = 1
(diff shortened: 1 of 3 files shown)


src/domains/observability/env/prod/terraform.tfvars
```diff
@@ -259,13 +259,31 @@ ehns_metric_alerts = {
       }
     ],
   },
+  no_wallet_ingestion_alert = {
+    aggregation = "Total"
+    metric_name = "IncomingMessages"
+    description = "Payment wallet onboarding written events less than 2000 detected in the last 24h"
+    operator    = "LessThanOrEqual"
+    threshold   = 2000
```
infantesimone, line 267: This threshold is too low; it is very close to the low-traffic onboarding number (see the traffic from November  2)

infantesimone (changes requested):

src/domains/observability/env/prod/terraform.tfvars
```diff
@@ -259,13 +259,31 @@ ehns_metric_alerts = {
...
+    frequency   = "PT1H"
+    window_size = "P1D"
+    dimension = [
+      {
+        name     = "EntityName"
+        operator = "Include"
+        values = [
+          "payment-wallet-evt-tx",
```
CianoDanilo, line 275: Are we sure this is the right entity? The azure dashboard works on the event hub name while filtering per entity name, payment-wallet-ingestion-dl

src/domains/observability/env/prod/terraform.tfvars
pietro-tota: Yes, maybe we can lower to 1000 in one day and see how it will go in prod after activating it

pietro-tota: Lowered to 1000 with https://github.com/pagopa/pagopa-infra/pull/2608/commits/23ab4ea37e9438c30995ed3a22cd5a59d8a690ec

src/domains/observability/env/prod/terraform.tfvars
pietro-tota: I double checked against metrics and seems you're right.
Updated values to use the eventhub name with https://github.com/pagopa/pagopa-infra/pull/2608/commits/fafd8bf4ed63bc4810d8aa055a26ae4d30182207
This configuration was taken from other alerts configuration but i think that there is a misconfiguration on other alerts

CianoDanilo (approved):

infantesimone (dismissed):

pietro-tota:
Apply on hold: alerts will be activated once rollout in production have been completed

infantesimone (approved):

PR #2646, opened 2024-12-10
fix: CHK-3593 fix ecommerce/wallet queue alert query

ciuffagianluca:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-common/03_storage.tf
```diff
@@ -391,17 +384,100 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_transient_enqu
...
+  resource_group_name = azurerm_resource_group.storage_ecommerce_rg.name
+  location            = var.location
+
+  action {
+    action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id, azurerm_monitor_action_group.ecommerce_opsgenie[0].id]
+    email_subject          = "Email Header"
+    custom_webhook_payload = "{}"
+  }
+  data_source_id = module.ecommerce_storage_transient.id
+  description    = format("Enqueuing rate for queue %s > ${each.value.threshold} during last ${each.value.time_window} minutes", replace("${each.value.queue_key}", "-", " "))
```
EmanueleBVtech, line 448: Since in the description field we are already using format, maybe we can avoid interpolation. Just an idea:

```description = format("Enqueuing rate for queue %s > %d during last %d minutes", replace(each.value.queue_key, "-", " "), each.value.threshold, each.value.time_window)
`

Comment (ciuffagianluca) [reply src/domains/ecommerce-common/03_storage.tf:448]: Yes, it's a good idea, but do you mind if we face this point in a dedicated task? If I do it here, I should do it for all alert, and it would be out of the scope. What do you think?

Comment (EmanueleBVtech) [reply src/domains/ecommerce-common/03_storage.tf:448]: yes of course, mine was just an idea, go ahead

**Review by EmanueleBVtech — APPROVED** [review]

File: src/domains/ecommerce-common/03_storage.tf [hunk 2]
```diff
@@ -391,17 +384,100 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_transient_enqu
...
+      "severity"    = 1
+      "time_window" = 15
+      "frequency"   = 15
+      "threshold"   = 40
+    },
+  ] : []
+}
+
+# Queue size: Ecommerce - ecommerce queues enqueues rate alert
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_expiration_enqueue_rate_alert" {
```
pietro-tota, line 436:
```suggestion
resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_enqueue_rate_alert_visibility_timeout_diff" {
```

make this alert name general

src/domains/ecommerce-common/03_storage.tf
```diff
@@ -391,17 +384,100 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_transient_enqu
...
+  trigger {
+    operator  = "GreaterThan"
+    threshold = 0
+  }
+}
+
+locals {
+  queue_expiration_alert_props = var.env_short == "p" ? [
+    {
+      "queue_key"   = "transactions-expiration-queue"
```
pietro-tota, line 426: we should add here also the other queues were events are written with a visibility timeout other than expiration queue.
here should be add also:
- all the \*retry\* queues (such as notifications-service-retry-queue, transaction-notifications-retry-queue and so on)
- transaction-auth-requested-queue (for NPG transactions event is re-written into visibility timeout to trigger GET state api call)
- transaction-auth-outcome-waiting-queue

all the above mentioned queues have events written with visibility timeout that could cause put - delete count to fail for a determined time slot

ciuffagianluca: You're right. I'm going to add those queues to this alert group

ciuffagianluca: Done here d34f9d6df049659827b7e46df82da0bde826a8f5

src/domains/ecommerce-common/03_storage.tf
```diff
@@ -302,61 +302,119 @@ resource "azurerm_monitor_diagnostic_setting" "ecommerce_transient_queue_diagnos
...
+    | where Diff > ${each.value.threshold}
+    QUERY
+    , "/${module.ecommerce_storage_transient.name}/${local.project}-${each.value.queue_key}"
+  )
+  severity    = each.value.severity
+  frequency   = each.value.frequency
+  time_window = each.value.time_window
+  trigger {
+    operator  = "GreaterThan"
+    threshold = 0
```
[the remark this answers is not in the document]
ciuffagianluca: This kind of threshold is related to the cardinality of the result set. In other words, you are saying that if you have one result from your query, you have to trigger the alert. The query itself filters results for the ones that are greater than the queue properties threshold.  So if you have at least one result, you have to trigger the alert.

pietro-tota (approved):
lgtm

src/domains/ecommerce-common/03_storage.tf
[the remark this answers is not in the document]
CianoDanilo: My fear is that either the query threshold or the trigger threshold is too low leading to trigger false positives but if we have already fine tuned this values and we're confident about them we can resolve this conversation.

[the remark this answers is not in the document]
ciuffagianluca: The second trigger threshold (this one highlighted) must be 0. 
The params threshold could be higher or lower than its value as it is set now.
I proceed to resolve this conversation.

EmanueleBVtech (dismissed):

src/domains/ecommerce-common/03_storage.tf
```diff
@@ -302,70 +302,128 @@ resource "azurerm_monitor_diagnostic_setting" "ecommerce_transient_queue_diagnos
...
+# Queue size: Ecommerce - ecommerce queues enqueues rate alert
+resource "azurerm_monitor_scheduled_query_rules_alert" "ecommerce_transient_enqueue_rate_alert" {
+  for_each            = { for q in local.queue_transient_alert_props : q.queue_key => q }
+  name                = "${local.project}-${each.value.queue_key}-rate-alert"
+  resource_group_name = azurerm_resource_group.storage_ecommerce_rg.name
+  location            = var.location
+
+  action {
+    action_group           = [data.azurerm_monitor_action_group.email.id, data.azurerm_monitor_action_group.slack.id, azurerm_monitor_action_group.ecommerce_opsgenie[0].id]
+    email_subject          = "Email Header"
```
pietro-tota, line 337: Change this email_subject field to reflect the alert reason.
Please change it also for the below alert too

pietro-tota (dismissed):
Well done, just few trivial issues

pietro-tota (dismissed):

src/domains/pay-wallet-common/03_storage.tf
```diff
@@ -182,17 +174,102 @@ resource "azurerm_monitor_scheduled_query_rules_alert" "pay_wallet_enqueue_rate_
...
+  action {
+    action_group           = each.value.action_group
+    email_subject          = "[pay-wallet] Enqueue rate for wallet queue too high"
+    custom_webhook_payload = "{}"
+  }
+  data_source_id = module.pay_wallet_storage[0].id
+  description    = format("Enqueuing rate for queue %s > ${each.value.threshold} during last ${each.value.time_window} minutes", replace("${each.value.queue_key}", "-", " "))
+  enabled        = true
+  query = format(<<-QUERY
+    let endDelete = datetime_local_to_utc(startofday(now()), 'Europe/Rome');
```
ciuffagianluca, line 243:
```suggestion
    let endDelete = datetime_local_to_utc(now(), 'Europe/Rome');
```

PR #2654, opened 2024-12-12
feat(paymen-wallet): PPABV-60 add new healthcheck for payment wallet

infantesimone:
### List of changes

- add new api to support healhtcheck for `payment-wallet`
### Motivation and context

This PR introduces a new API endpoint to provide a liveness check for the `payment-wallet`. The purpose of this endpoint is to monitor whether the service is operational and capable of handling requests. For the moment, the focus is solely on liveness. In future iterations, readiness checks will be explored to assess the service's ability to handle traffic after initialization and dependency availability.
This API will initially be integrated into `StatusPage`.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/pay-wallet-app/api/healthcheck/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,71 @@
+<policies>
+    <inbound>
+        <base />
+
+        <send-request mode="new" response-variable-name="walletServiceLiveness" timeout="10">
```
pietro-tota, line 5: maybe here we can add an "ignore-error" set to true for all of this send-request to avoid that the whole request results in an error if one of those modules is down and then add a check against http error code too in the outbound policy section.
This way the response will successfully be returned even if a single service is down (not responding to healthcheck api call)

pietro-tota (dismissed):
🚀  just a consideration about responses error handling

src/domains/pay-wallet-app/api/healthcheck/v1/_base_policy.xml.tpl
infantesimone: correct!

PR #2717, opened 2025-01-14
feat: CHK-2894 remove onetrust from checkout fe CSP

ciuffagianluca:
Depends on https://github.com/pagopa/pagopa-checkout-fe/pull/414

This CSP modification could be applied AFTER dependent Checkout PR merge that removes one trust usage for prod env

Remove onetrust domain from CSP rules

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


scaminati-bv (approved):

pietro-tota (approved):
LGMT

PR #2737, opened 2025-01-22
feat: [PAGOPA-2527] Refactoring FdR-Fase3 collections

andrea-deri:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/fdr-common/03_cosmos_mongodb_fdr.tf
```diff
@@ -63,105 +63,42 @@ resource "azurerm_cosmosdb_mongo_database" "fdr" {
...
-    {
-      name = "fdr_payment_history"
-      indexes = [{
-        keys   = ["_id"] # index + ref_fdr_reporting_flow_name + ref_fdr_id
-        unique = true
         }
       ]
-      shard_key = null
+      shard_key   = null,
+      ttl_seconds = "3024000" # 30 days + 5 days (deltaTime)
```
cap-ang, line 82: Could be useful export it as variable?
e.g. for different environment can be useful have different TTL

src/domains/fdr-common/03_cosmos_mongodb_fdr.tf
```diff
@@ -63,105 +63,42 @@ resource "azurerm_cosmosdb_mongo_database" "fdr" {
...
-          unique = false
-        }
+        #{
+        #  keys   = ["ref_fdr.sender_psp_id", "iuv", "created"] # payment_by_iuv_idx
+        #  unique = false
+        #}
       ]
-      shard_key = null
+      shard_key   = null,
+      ttl_seconds = "3024000" # 30 days + 5 days (deltaTime)
```
cap-ang, line 101: ditto

src/domains/fdr-common/03_cosmos_mongodb_fdr.tf
```diff
@@ -63,105 +63,42 @@ resource "azurerm_cosmosdb_mongo_database" "fdr" {
 locals {
```
cap-ang, line 63: Maybe there is a missing update of the description
https://github.com/pagopa/pagopa-infra/blob/6e9b7d892ddd8d5cbddc8340d23e648669e9e8d9/src/domains/fdr-common/03_cosmos_mongodb_fdr.tf#L48

andrea-deri: My bad, just a typo. It will be removed on the next commit.

src/domains/fdr-common/03_cosmos_mongodb_fdr.tf
andrea-deri: Good idea, I will set TTL values as environment variables in the next commit.

cap-ang (changes requested):
Good work!

Some small improvements that could be made
- you may want to store `ttl_seconds` value in a variable (`var.container_ttl`) for easier modification across different environments
- if the queries don’t require sorting by all fields, consider breaking the compound indexes into single-field indexes to optimize indexing costs (as described in the comment)
- it may be helpful to document why the shard key is set to null

andrea-deri:
@cap-ang trying to answering your points:
 1. I've updated the TTL values defining a dedicated variable, as you correctly exposed
 2. That's correct, as described in the CosmosDB documentation, but there are some "issues" related to that:
   2.1. Some of them are unique index that provide DB-level constraints on data and are also useful for indexed search queries 
   2.2. if we made different indexes, we could possibly not optimize certain sort operation that could be useful on "paginated read" APIs
   2.3. IMHO, the most important pro of MongoDB APIs in this (and other) project is the possibility to deploy the DB on another CSP, avoiding a vendor lock-in. CosmosDB's implementation is optimized on defining indexes for each document field but this could not be true on another CSP's implementation. Due to the fact that the read performance are equals, I would follow this "generic" way.   
 3. In FdR case, the sharding is very bad: for example, on `fdr_payment` the key that could be used as `shard_key` (`ref_fdr.id`) could not be so sparse and this can lead to unbalanced shards and bad-performing queries (and in a read-operation of 100k documents this can be fatal). Choosing another field as `shard_key` does not permits to use multiple-document transactions in insert operation because in MongoDB that could be performed only on the same shard.

cap-ang (approved):
LGTM 👍 

Regarding the sharding strategy, I understood this: it does not improve performance since the queries we will make will not use the eligible shard key. Correct me if I have misread

andrea-deri:
Correct! 
But it can also lead to transactions being impossible, because document on different shards cannot be part of the same transaction.

aomegax (approved):

PR #2739, opened 2025-01-22
feat: [CHK-3649] set buyerbanks storage access_tier to Cool

scaminati-bv:
### List of changes
- set **Cool** on access_tier of buyerbanks storage

### Motivation and context
The goal of this PR is to set buyerbank storage in a cool status to reduce costs since that the related function is dismissed and this storage is not used anymore.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [ ] No

### Other information

---


src/core/buyerbanks_function.tf
```diff
@@ -187,7 +187,7 @@ module "buyerbanks_storage" {
   account_kind               = "StorageV2"
   account_tier               = "Standard"
   account_replication_type   = var.buyer_banks_storage_account_replication_type
-  access_tier                = "Hot"
+  access_tier                = "Cool"
```
pietro-tota, line 190: why set it to cool storage and not to [archive one](https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview#archive-access-tier)?

scaminati-bv: Because the only available options are **Hot** and **Cool** . This is the [documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#access_tier-1).

CianoDanilo (approved):

pietro-tota (approved):

PR #2776, opened 2025-02-10
chore: [PAGOPA-2631] Fix index DB FdR

aomegax:
### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


aomegax:
> @aomegax why does't we shift this part on pagopa-fdr ?
> 
> I just have one doubt on our pgflex instance `pagopa-<ENB>-weu-fdr-flexible-postgresql` we'll have two db. image
> 
> So both FDR1 and FDR3 will have this part of migration schema in their repositories
> 
> do you agree?
> 
> cc @pagopa/pagopa-team-core

Yes, sure, but I would create a specific task to carry out the change, it means to change pipeline ref on `pagopa-azure-devops`, to move liquibase to `pagopa-fdr-nodo-dei-pagamenti` and so on. Not a quick job.

pasqualespica (approved):

pasqualespica:
@aomegax we have a dedicate action to apply it into PROD env , can we merge it? since it is already in DEV and UAT env

pasqualespica:
> > @aomegax why does't we shift this part on pagopa-fdr ?
> > I just have one doubt on our pgflex instance `pagopa-<ENB>-weu-fdr-flexible-postgresql` we'll have two db. image
> > So both FDR1 and FDR3 will have this part of migration schema in their repositories
> > do you agree?
> > cc @pagopa/pagopa-team-core
> 
> Yes, sure, but I would create a specific task to carry out the change, it means to change pipeline ref on `pagopa-azure-devops`, to move liquibase to `pagopa-fdr-nodo-dei-pagamenti` and so on. Not a quick job.

Ok it's right , Let's create a new task and maybe put it in the next sprint as ** Should Have**

PR #2791, opened 2025-02-13
chore(queue-rate-alert): increase dynamic threshold to 2%

pietro-tota:
### List of changes
Increase dynamic threshold to 2% for queue PUT vs DELETE alert in order to reduce false positive alert triggering

### Motivation and context
Increase dynamic threshold to 2% for queue PUT vs DELETE alert in order to reduce false positive alert triggering

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-common/03_storage.tf
```diff
@@ -315,15 +315,15 @@ locals {
       "time_window"       = 30
       "frequency"         = 15
       "threshold"         = 10
```
ciuffagianluca, line 317:
```suggestion
      "threshold"         = 20
```

PR #2800, opened 2025-02-17
fix: [PAGOPA-2671] updating OpenAPI for new FdR-Fase1 internal endpoint

andrea-deri:
As described in [this PR](https://github.com/pagopa/pagopa-fdr-nodo-dei-pagamenti/pull/76), this contains a new feature made for FdR-Fase1 application in order to handle the migration from Nodo dei Pagamenti's primitives to FdR's primitives and track Nexi's `nodoInviaFlussoRendicontazione`. This endpoint will be used only in transitory way, being dismissed when all traffic will switch on FdR-Fase1 application from Nodo dei Pagamenti. 

### List of changes
 -  Adding new API for register required information about Nexi-handled `nodoInviaFlussoRendicontazione`

### Motivation and context
These changes are required for transitory processes on NdP-to-FdR1 transition

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


andrea-deri:
After some in-deep analysis about the usage of the internal API, we found that it is not necessary to include the `register-for-validation` API in the "FdR1 Internal" API product. This because that API is used only by APIM policy (as defined in [this PR](https://github.com/pagopa/pagopa-infra/pull/2801)) and it is unnecessary to expose it to internal use for other FdR-related applications.  
I am closing this PR because it is not needed anymore.  
Thanks to @aomegax for clarification and to @pasqualespica for the review.

PR #2804, opened 2025-02-17
feat(checkout-common): [CHK-3695] add Redis Checkout cache instance

pietro-tota:
Note for apply:
1. first perform apply with target `azurerm_key_vault_secret.checkout_opsgenie_webhook_token` (create secret for Opsgenie webhook api key
2. update secret with right api key
3. flat apply to create all other Redis resources (Redis cache, subnet and alert)

### List of changes
This PR configure a Redis cache for Checkout domains that will be used by the checkout-auth service.
The following resources have been added:
- checkout opsgenie webhook key secret (used by the opsgenie action group for Redis cache alerts, only for prod env)
- checkout opsgenie action group (used by the opsgenie action group for Redis cache alerts, only for prod env)
- redis checkout resource group
- redis subnet 
- redis cache (version 6)
- redis memory usage alerts (only for prod env)

Redis cache tier is set to Basic (C0) for DEV and UAT envs and to Premium (P1) for PROD env (with zone replication enabled)

### Motivation and context
This Redis will be used by checkout-auth b.e. during authentication flow

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---

### DIFF
--- a/src/domains/checkout-common/03_redis.tf
+++ b/src/domains/checkout-common/03_redis.tf
@@ -98,11 +98,11 @@ module "pagopa_checkout_redis" {
   sku_name                      = var.redis_checkout_params.sku_name
   enable_authentication         = true
   redis_version                 = var.redis_checkout_params.version
-  public_network_access_enabled = var.env_short == "d"
+  public_network_access_enabled = false
   zones                         = var.redis_checkout_params.zones
 
   private_endpoint = {
-    enabled              = var.env_short != "d"
+    enabled              = true
     virtual_network_id   = "TODO"
     subnet_id            = module.pagopa_checkout_redis_snet.id
     private_dns_zone_ids = [data.azurerm_private_dns_zone.privatelink_redis_cache_windows_net.id]
--- a/src/domains/checkout-common/03_redis.tf
+++ b/src/domains/checkout-common/03_redis.tf
@@ -71,7 +71,7 @@ module "pagopa_proxy_redis" {
 resource "azurerm_resource_group" "redis_checkout_rg" {
   name     = "${local.project}-redis-rg"
   location = var.location
-  tags = var.tags
+  tags     = var.tags
 }
 
 # Redis cache subnet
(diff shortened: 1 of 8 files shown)


src/domains/checkout-common/env/weu-uat/terraform.tfvars
```diff
@@ -40,3 +40,13 @@ pagopa_proxy_redis_family   = "C"
 redis_private_endpoint_enabled = true
 redis_zones                    = []
 
+# Checkout Redis
+redis_checkout_params = {
+  capacity   = 0
+  sku_name   = "Basic"
+  family     = "C"
+  version    = 6
```
infantesimone, line 48: Have you considered using Redis 7?
https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-how-to-upgrade

pietro-tota: It seems that version 7 is available only for Enterprise tier and not for Basic/Standard/Premium ones.
Moreover actually used module only accept 4 and 6 as version, so maybe we should upgrade module first too
Should we use Enterprise version?

ciuffagianluca (dismissed):
🚀 well done!

src/domains/checkout-common/03_redis.tf
```diff
@@ -53,3 +57,130 @@ module "pagopa_proxy_redis" {
...
+  name     = "${local.project}-redis-rg"
+  location = var.location
+  tags     = var.tags
+}
+
+# Redis cache subnet
+module "pagopa_checkout_redis_snet" {
+  source = "git::https://github.com/pagopa/terraform-azurerm-v3.git//subnet?ref=v8.83.1"
+
+  name                                      = format("%s-redis-snet", local.project)
```
infantesimone, line 81: You can use interpolation as it is used in the definition of the resource group

pietro-tota: fixed with https://github.com/pagopa/pagopa-infra/pull/2804/commits/018e75896345fd1feb4e22e34c577ef837fa310b

infantesimone (approved):

EmanueleBVtech (approved):

PR #2848, opened 2025-03-03
fix: update bizevents-service openapi

a-mannia:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


a-mannia:
Skipped because it has already been implemented [here](https://github.com/pagopa/pagopa-infra/pull/2814).

PR #2880, opened 2025-03-13
feat(checkout-auth): [CHK-3834] add checkout v3 policies

CianoDanilo:
### List of changes
Added checkout v3 policies for POST /auth-request and GET /transaction

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


pietro-tota:
Not needed, we will just update v1 and v2 apis version to inject userId without any modification to externally exposed apis

PR #2888, opened 2025-03-17
feat(checkout auth): alerts for checkout auth api

pietro-tota:
### List of changes
Add checkout authenticated api's availability alerts.
There have been created 4 alerts that are:
- one for internal api -> this alert monitors checkout-auth service api that have no external downstream dependency
- one for external api -> this alert monitors checkout-auth service api that have external downstream dependency (One identity in this case)
- one for unauthorized percentage -> this alert monitors unauthorized api calls percentage vs total api calls in order to be triggered once the unauthorized percentage is greater then a threshold percentage

There have been modified also _ecommmerce-for-checkout_ availability alert that already includes all versions api to correctly detect payment-requests v3 request too)

### Motivation and context
Those modifications are needed in order to add alerts that will be triggered if availability is lower than defined SLO, allowing to perform intervention to restore service

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


scaminati-bv (approved):

src/domains/checkout-common/00_alerts.tf
```diff
@@ -0,0 +1,154 @@
...
+    email_subject          = "[Checkout] Auth service internal API availability alert"
+    custom_webhook_payload = "{}"
+  }
+  data_source_id = data.azurerm_api_management.apim.id
+  description    = "Checkout-auth-service internal API availability less than or equal 99% in the last 30 minutes"
+  enabled        = true
+  query = (<<-QUERY
+AzureDiagnostics
+| where url_s startswith 'https://api.platform.pagopa.it/checkout/auth-service/v1'
+  and operationId_s != "authenticateWithAuthToken" //exclude POST auth/token api
```
CianoDanilo, line 27: I think we could avoid filtering for operationId_s, POST /auth/token API won't be called by the frontend and it is not a public API, do you agree?

src/domains/checkout-common/00_alerts.tf
```diff
@@ -0,0 +1,154 @@
...
+  data_source_id = data.azurerm_api_management.apim.id
+  description    = "Checkout-auth-service external API availability less than or equal 95% in the last 30 minutes"
+  enabled        = true
+  query = (<<-QUERY
+AzureDiagnostics
+| where url_s startswith 'https://api.platform.pagopa.it/checkout/auth-service/v1'
+  and operationId_s == "authenticateWithAuthToken" //include only POST auth/token api
+| summarize
+    Total=count(),
+    Success=countif(responseCode_d < 500 and DurationMs < 500)
```
CianoDanilo, line 67: Do we have to tune the duration here or we're already confident about this value?

src/domains/checkout-common/00_alerts.tf
```diff
@@ -0,0 +1,154 @@
...
+  }
+  data_source_id = data.azurerm_api_management.apim.id
+  description    = "Checkout-auth-service login success rate less than or equal 95% in the last 30 minutes"
+  enabled        = true
+  query = (<<-QUERY
+
+AzureDiagnostics
+| where url_s startswith 'https://api.platform.pagopa.it/checkout/auth-service/v1'
+| summarize
+GetLoginCount=countif(responseCode_d == 200 and operationId_s == "authLogin"),
```
CianoDanilo, line 103: Can we tie this count to /login and /token with same requestId/correlationId and count /token with response code != 200? I think this query could lead to false positives in case many users click on login button but then decide not to proceed further.

src/domains/ecommerce-common/00_alerts.tf
```diff
@@ -58,7 +58,7 @@ AzureDiagnostics
 | where url_s startswith 'https://api.platform.pagopa.it/ecommerce/checkout/'
 | summarize
     Total=count(),
-    Success=countif(responseCode_d < 500 or url_s startswith "https://api.platform.pagopa.it/ecommerce/checkout/v1/payment-requests" and ( responseCode_d == 502 or responseCode_d == 504))
+    Success=countif(responseCode_d < 500 or (operationId_s in ("getPaymentRequestInfo", "getPaymentRequestInfoV3") and responseCode_d in (502, 504)))
```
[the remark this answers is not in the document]
pietro-tota: Yes since here we are modelling all authenticated flow api alerts (for runbook etc.)
V3 apis have been added for authenticated flow but no modification to existing alert were performed.
This modification take into account this aspect too

src/domains/checkout-common/00_alerts.tf
pietro-tota: Yes, this is true. I modeled this alert as per current implementation where a correlation/request id is not present. 
I think that we can leave it as is for now (just to have at least a counting alert) and modify query in the correlation task. what do you think?

CianoDanilo (approved):

EmanueleBVtech (approved):

scaminati-bv (approved):

PR #2892, opened 2025-03-18
[CHK-3816] feat: manage x-rpt-id header into auth-service api

scaminati-bv:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl
+++ b/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl
@@ -31,6 +31,15 @@
         "description": "GET login endpoint with reCAPTCHA code",
         "parameters": [
           {
+            "in": "header",
+            "name": "x-rpt-id",
+            "required": false,
+            "description": "Optional RPTID",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
             "in": "query",
             "name": "recaptcha",
             "required": true,
@@ -92,6 +101,17 @@
         "operationId": "authUsers",
         "summary": "Get user information",
         "description": "GET user information",
+        "parameters": [
+          {
+            "in": "header",
+            "name": "x-rpt-id",
+            "required": false,
+            "description": "Optional RPTID",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
         "security": [
           {
             "bearerAuth": []
@@ -159,6 +179,17 @@
         "operationId": "authLogout",
         "summary": "Logout endpoint",
         "description": "POST logout endpoint",
+        "parameters": [
+          {
+            "in": "header",
+            "name": "x-rpt-id",
+            "required": false,
+            "description": "Optional RPTID",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
         "security": [
           {
             "bearerAuth": []
@@ -209,6 +240,17 @@
         "operationId": "authenticateWithAuthToken",
         "summary": "Authentication endpoint",
         "description": "POST authentication endpoint with auth code",
+        "parameters": [
+          {
+            "in": "header",
+            "name": "x-rpt-id",
+            "required": false,
+            "description": "Optional RPTID",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
         "requestBody": {
           "required": true,
           "content": {


src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl
```diff
@@ -30,6 +30,15 @@
         "summary": "Login endpoint",
         "description": "GET login endpoint with reCAPTCHA code",
         "parameters": [
+          {
+            "in": "header",
+            "name": "x-rpt-id",
+            "required": false,
+            "description": "Optional RPTID",
```
pietro-tota, line 37:
```suggestion
            "description": "Optional RPT ID",
```

src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl
```diff
@@ -337,10 +379,6 @@
       "UserInfoResponse": {
         "type": "object",
         "properties": {
-          "userId": {
```
pietro-tota, line 340: why removing here the userId field?
This is object for another PR, please restore here

src/domains/ecommerce-app/api/ecommerce-checkout/v3/_transaction_policy.xml.tpl
```diff
@@ -1,14 +1,32 @@
...
+            foreach (JObject notice in ((JArray)(context.Variables["paymentNotices"]))) {
+                if(result != "") {
+                    result += ", ";
+                }
+                if( notice.ContainsKey("rptId") == true ){
+                    result += notice["rptId"].Value<string>();
+                }
+            }
+            return result;
+        }" />
```
pietro-tota, line 18:
```suggestion
  <set-variable name="rptIds" value="@{
            string result = "";
            JArray paymentNotices = ((JArray)(context.Variables["paymentNotices"]));
            if(paymentNotices != null){
                List<string> rptIds = paymentNotices.Select(notice => notice["rptId"]?.Value<string>()).ToList();
                result =  String.Join(",",rptIds);
            }
            return result;
        }" />
```

this could be simplified as follow

scaminati-bv: Yes much better.

src/domains/ecommerce-app/api/ecommerce-checkout/v3/_base_policy.xml.tpl
```diff
@@ -52,11 +52,14 @@
...
-          <set-header name="Authorization" exists-action="override">
-              <value>@("Bearer " + (string)context.Variables["authToken"])</value>
-          </set-header>
+            <set-url>@($"https://${checkout_ingress_hostname}/pagopa-checkout-auth-service/auth/validate")</set-url>
+            <set-method>GET</set-method>
+            <set-header name="Authorization" exists-action="override">
+                <value>@("Bearer " + (string)context.Variables["authToken"])</value>
+            </set-header>
+            <set-header name="x-rpt-ids" exists-action="override">
+              <value>@((string)context.Variables.GetValueOrDefault("rptId",""))</value>
```
CianoDanilo, line 61: Where is the variable set? Should we use context.Request.Headers.GetValueOrDefault instead?

scaminati-bv: This variabile is setted [here](https://github.com/pagopa/pagopa-infra/blob/main/src/domains/ecommerce-app/api/ecommerce-checkout/v3/_payment_request_policy.xml.tpl). I don't change this because is already used in this part.

CianoDanilo: Aren't we adding rptId to every API we can? I mean, maybe as it is now, payment-request is the only API which has it but, client side, we could pass it to every other APIs in order to propagate it to auth-service, right? Maybe I got the task definition wrong? @pietro-tota @infantesimone

scaminati-bv: In addition to the payment request, the post transaction also has the rpId read and used in a specific [policy](https://github.com/pagopa/pagopa-infra/blob/main/src/domains/ecommerce-app/api/ecommerce-checkout/v3/_transaction_policy.xml.tpl). 
The task currently refers to all api calls to auth-service, but if we want i can integrate payment-method apis with the same header.

scaminati-bv: After discussion we decided to leave the header in the singular version and it was decided to add the header x-rpt-id to the v3 checkout api as well.

src/domains/ecommerce-app/api/ecommerce-checkout/v3/_transaction_policy.xml.tpl
```diff
@@ -49,21 +52,10 @@
 
         <!-- pass rptId value into header START -->
         <set-header name="x-rpt-id" exists-action="delete" />
```
CianoDanilo, line 54: Should we remove this line? Since we are using context.Request.Headers.GetValueOrDefault method later in the policy, would it be a problem deleting the header before?

scaminati-bv: Done here bc0304928762546f99c854497b421f2ac859a90b

scaminati-bv: In this commit 4100553ba7aee9dab999f8bc5df2f67eadb66e1b I optimized header delete condition.

scaminati-bv: In this commit 4100553ba7aee9dab999f8bc5df2f67eadb66e1b I optimized header delete condition.

CianoDanilo (approved):

pietro-tota (approved):

PR #2913, opened 2025-03-24
feat(auth): [CHK-3874] update get users api

CianoDanilo:
### List of changes
Removed user fiscal code from getUsers API response body
Added validation against openAPI spec to getUsers API response body

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_get_user_filter.xml.tpl
```diff
@@ -0,0 +1,43 @@
...
+                <set-header name="Content-Type" exists-action="override">
+                    <value>application/json</value>
+                </set-header>
+                <set-body>@{
+                    return new JObject(
+                            new JProperty("name", ((JObject) context.Variables["body"])["name"]),
+                            new JProperty("familyName", ((JObject) context.Variables["body"])["familyName"])
+                            ).ToString();
+                        }</set-body>
+                <validate-content unspecified-content-type-action="prevent" max-size="512000" size-exceeded-action="detect" errors-variable-name="responseBodyValidation">
```
pietro-tota, line 31: 🚲 
maybe 512KB as max response size makes size check useless, i would set it to 1K
what do you think?

pietro-tota (approved):
LGMT just a trivial issue

src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_get_user_filter.xml.tpl
CianoDanilo: Unfortunately it seems max-size is a required field :(. Maybe we can lower it to 128KB? I'm afraid 1KB max size would always trigger the detection-mode, what do you think?

pietro-tota: consider that the mocked response is 70 byte length:
``` 
{"userId":"fiscalNumber","name":"NomeTest","familyName":"CognomeTest"}
```
(just checked the Content-length response header)
and here we are also removing user id part
why are you afraid that 1KB is not enough?  Maybe 1KB is still too high as value

CianoDanilo: Ok then I'll set it to 1KB

CianoDanilo: Done in [5ac8441](https://github.com/pagopa/pagopa-infra/pull/2913/commits/5ac8441e3f28f91d975869a60e643b6025f45fba)

pietro-tota (approved):

scaminati-bv (approved):

infantesimone (approved):

PR #2960, opened 2025-04-09
feat: adding v2 for transaction auth requests service api

sfarralorenzo:
Adding v2 for the transaction auth requests service api

### Motivation and context
We are transitioning from our current flow, based on base64-encoded transaction IDs, to plain transaction IDs.

The `v1` system will remain unchanged in both syntax and functionality.
We're introducing a `v2` for the plain transaction ID format, and both versions can operate independently to avoid disruption.
This separation ensures backward compatibility while we implement the updated approach.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

Targets:
 - `azurerm_api_management_api_operation_policy.auth_request_gateway_policy_v2`
 - `module.apim_ecommerce_transaction_auth_requests_service_api_v2`
 - `azurerm_key_vault_secret.transactions_service_auth_update_api_key_v2`

4 changes expected

---


pietro-tota (dismissed):
🚀

EmanueleBVtech (dismissed):

EmanueleBVtech (dismissed):

pietro-tota (approved):

pietro-tota (approved):

PR #2962, opened 2025-04-09
feat: use plain transaction IDs in NPG notifications and redirect flows

sfarralorenzo:
Depends on https://github.com/pagopa/pagopa-infra/pull/2960 .
remove base64 encoding of transaction IDs and upgrade to v2 API of our patch call to the transaction-service auth-requests  api.

### List of changes

For redirect and npg notifications flow:

- remove base64 encoding transformation of transaction IDs 
- update API endpoint to use v2 version of transaction-service auth-requests
- Use transaction IDs directly in API calls without transformation

### Motivation and context

This intervention was originally made for the management of XPAY on PGS, but now it's possible to remove this differentiation and use the transaction ID as-is in all calls to the transactions-service.
This would also make it easier to correlate all calls belonging to the same transaction.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information

---

targets:
 - `resource.azurerm_api_management_api_operation_policy.npg_notifications_policy`
 
 - `module.apim_ecommerce_redirect_outcome_api_v1`


pietro-tota (dismissed):
🚀

EmanueleBVtech (dismissed):

EmanueleBVtech (approved):

pietro-tota (approved):

PR #2963, opened 2025-04-10
feat: [CHK-4046] add userId index into transactions-view document

scaminati-bv:
### List of changes
- set userId index into transactions-view document

### Motivation and context
The goal of this PR is to add index that allows helpdesk to search for transactions by fiscal code.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/ecommerce-common/03_cosmosdb.tf
```diff
@@ -144,6 +144,10 @@ locals {
         {
           keys   = ["email.data"]
           unique = false
+        },
+        {
+          keys   = ["userId"]
+          unique = false
```
EmanueleBVtech, line 150: userId is an optional field that may be missing in the scenario of a guest flow. Please fix this index to skip indexing when the field is missing.

```suggestion
          unique = false,
          condition = "exists(userId)"  # Add this condition to handle missing value
```

scaminati-bv: By default Text index is a Sparse index [MongoDb doc](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-text/#sparse-property). This means that by default possible null values are already handled.

EmanueleBVtech: great!

src/domains/ecommerce-common/03_cosmosdb.tf
```diff
@@ -144,6 +144,10 @@ locals {
         {
           keys   = ["email.data"]
           unique = false
+        },
+        {
+          keys   = ["userId"]
```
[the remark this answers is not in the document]
scaminati-bv: As you can see on the [common](https://github.com/pagopa/pagopa-ecommerce-commons/blob/main/src/main/java/it/pagopa/ecommerce/commons/documents/v2/Transaction.java) repository userId is a string field.
Email is a Object with data field, so to read value `email.data` is necessary.

EmanueleBVtech, line 149: this index is also required on the eventstore collection, can you replicate ?

scaminati-bv: Why is required into eventstore? Helpdesk search should only access transaction-view.

EmanueleBVtech: mb, not necessary

EmanueleBVtech (approved):

infantesimone (approved):

pietro-tota (approved):

PR #2982, opened 2025-04-18
[PIDM-451] fix: renamed header from x-rpt-id to x-rpt-ids

FrancescoDiTonnoJakala:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_base_policy.xml.tpl
+++ b/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_base_policy.xml.tpl
@@ -20,7 +20,7 @@
         <allowed-headers>
           <header>Content-Type</header>
           <header>Authorization</header>
-          <header>x-rpt-id</header>
+          <header>x-rpt-ids</header>
         </allowed-headers>
       </cors>
       <!-- Feature flag check - is authentication enabled -->
--- a/src/domains/ecommerce-app/api/ecommerce-checkout/v3/_base_policy.xml.tpl
+++ b/src/domains/ecommerce-app/api/ecommerce-checkout/v3/_base_policy.xml.tpl
@@ -17,7 +17,7 @@
           <header>lang</header>
           <header>x-correlation-id</header>
           <header>x-client-id-from-client</header>
-          <header>x-rpt-id</header>
+          <header>x-rpt-ids</header>
         </allowed-headers>
       </cors>
       <base />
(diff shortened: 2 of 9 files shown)


src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_base_policy.xml.tpl
```diff
@@ -20,7 +20,7 @@
         <allowed-headers>
           <header>Content-Type</header>
           <header>Authorization</header>
-          <header>x-rpt-id</header>
+          <header>x-rpt-ids</header>
```
pietro-tota, line 23: we cannot modify header this way: this will cause CORS errors for older checkout versions.
here you have to allow both headers, x-rpt-id and x-rpt-ids to allow accepting both version.

After deployment have been processed successfully the old x-rpt-id header can be removed from allowed-headers section. I would wait for a day or two, just to be sure there is no cached version still up & running

pietro-tota (changes requested):

src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-infra/pull/2982/commits/6c2ec75c049f94410a8f6a8cbdbf22a04b775732

pietro-tota: this is not the only affected point, look at all the base policies with cors section please. i just pin out this file for explain the problem

src/domains/ecommerce-app/api/ecommerce-checkout/v3/_base_policy.xml.tpl
```diff
@@ -17,7 +17,7 @@
           <header>lang</header>
           <header>x-correlation-id</header>
           <header>x-client-id-from-client</header>
-          <header>x-rpt-id</header>
+          <header>x-rpt-ids</header>
```
pietro-tota, line 20: ditto

FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-infra/pull/2982/commits/bddfb42a52619f7e4eb7499ff8caafaa54f24548

pietro-tota (approved):

PR #2993, opened 2025-04-23
feat(aca): add WL identity initialization

pietro-tota:
### List of changes
- create WL identity initialization in common domain
- update all modules in order to correctly use the latest v3 module
- create WL identity federation in app domain
- update middleware to use WL identity
 

### Motivation and context
Those modifications are required to update aca applications to use WL identities in place of pod identity

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
🚀

PR #3009, opened 2025-04-29
feat: PIDM-450 add alert and update openapi

cap-ang:
### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/gps-app/00_alert_gpd_core.tf
```diff
@@ -179,3 +179,37 @@ AzureDiagnostics
...
+    email_subject          = "GPD-SEND-STATUS"
+    custom_webhook_payload = "{}"
+  }
+
+  data_source_id = data.azurerm_application_insights.application_insights.id
+  description    = "An exception occurred when GPD tries to call SEND for updating notifications, it does NOT cause failures to asynchronous payments."
+  enabled        = true
+  query = (<<-QUERY
+let threshold = 0.99;
+traces
```
FedericoRuzzier, line 201: Why traces instead of exceptions?

cap-ang: Because the log added [here](https://github.com/pagopa/pagopa-debt-position/pull/301) are saved as traces. We catch the exception when communication with SEND fail.

FedericoRuzzier (dismissed):
LGTM 👍

pasqualespica (dismissed):

PR #3019, opened 2025-05-06
chore: adding wl identity apiconfig PIDM-487

FedericoRuzzier:
### List of changes
Adding wl identity for api-config domain
Target:

```
- apiconfig-common: -target=module.identity_cd_01 -target=null_resource.github_runner_app_permissions_to_namespace_cd_01 -target=module.workload_identity_init -target="module.api_config_ica_sa.azurerm_advanced_threat_protection.this"

- apiconfig-app: -target=module.workload_identity_configuration -target=module.tls_checker -target=helm_release.cert_mounter -target=module.cert_mounter -target=module.gh_runner_job
 ```

### Motivation and context
Needed in order to proceed with helm upgrade and switch to workload identity

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/apiconfig-common/10_github_identity.tf
sfarralorenzo, line 1: I think we should add the other repos in the `repos_01` list (specifically `pagopa-api-config` which is currently in WIP) because could it be the root cause of the volume mount failure in the pod

pasqualespica (approved):

PR #3037, opened 2025-05-15
feat(jwt-token-issuer): [CHK-4130] add certificate with rotation policy

pietro-tota:
### List of changes
Add certificate that will be used by jwt-token-issuer-service to retrieve both public and private RSA keys, used to sign JWT tokens.
Generated certificate will be valid for 1 month with automatic certificate rotation 2 days before expiration

### Motivation and context
This azure resource leverage service from public/private keys generation and rotations.
Certificate have been used in place of RSA key since azure does not allow to export private keys part but for certificate yes, allowing to use any library for jwt token signature.

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-common/02_security.tf
```diff
@@ -704,4 +704,44 @@ resource "azurerm_key_vault_secret" "ecommerce_gha_bot_pat" {
...
+    secret_properties {
+      content_type = "application/x-pkcs12"
+    }
+
+    x509_certificate_properties {
+      key_usage = [
+        "digitalSignature"
+      ]
+      subject            = "CN=${var.env}-${var.domain}-jwt-issuer"
+      validity_in_months = 1
```
CianoDanilo, line 744: Could it be useful to set different validities to different envs?

CianoDanilo (dismissed):

ciuffagianluca (approved):

CianoDanilo (approved):

PR #3121, opened 2025-06-11
feat(api-key): [CHK-4178] add api key secrets for payment-requests service

CianoDanilo:
### List of changes

In this PR, we created the primary and secondary API keys and 3 secrets: two are injected into the backend (payment-requests) and contain the primary and secondary keys; the third secret is injected into the clients (as named values for the APIM policy) and contains the value of the currently active API key (the value points to one of the secrets containing the primary or secondary key, based on a Terraform variable). This setup should make key rotation easier.

For the domain where the API keys are defined (in this case, ecommerce), no manual copy/paste is needed to sync secrets and API keys — everything can be managed through Terraform.

For external domains that need the API keys, we’ve decided to keep Key Vaults segregated and replicate the secret containing the currently enabled API key to each domain that requires it. This introduces some manual management, as the new secret needs to be manually aligned with the “master” value from the domain where the API keys are actually created and rotated.
### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/ecommerce-common/02_security.tf
```diff
@@ -719,6 +719,34 @@ resource "azurerm_key_vault_secret" "service_management_opsgenie_webhook_token"
...
+
+resource "azurerm_key_vault_secret" "ecommerce_payment_requests_secondary_api_key" {
+  name         = "ecommerce-payment-requests-secondary-api-key"
+  value        = random_password.ecommerce_payment_requests_secondary_api_key_pass.result
+  key_vault_id = module.key_vault.id
+}
+
+resource "azurerm_key_vault_secret" "ecommerce_payment_requests_active_api_key" {
+  name         = "ecommerce-payment-requests-active-api-key"
+  value        = var.ecommerce_payment_requests_api_key_use_primary ? azurerm_key_vault_secret.ecommerce_payment_requests_primary_api_key.value : azurerm_key_vault_secret.ecommerce_payment_requests_secondary_api_key.value
```
infantesimone, line 746: so to activate the secondary is it necessary to apply terraform with consequent restart of pods that read that secret?

CianoDanilo:
> it's not clear to me the goal of `ecommerce_payment_requests_active_api_key`

The goal is to provide a stable reference point for clients that need to use the API key — without needing to know which key (primary or secondary) is currently active. It abstracts away whether the primary or secondary key is being used, making key rotation seamless for consumers.

src/domains/ecommerce-common/02_security.tf
infantesimone: Thanks! however, could we populate the named value directly with one of the two secrets? this way, in case of rotation, only a single apply would be needed to inject it into the named value (currently, we need to do one apply in app and then another in common)

infantesimone (approved):

infantesimone (approved):

PR #3163, opened 2025-06-20
feat: CHK-4202 handle calculate fees v2 for io get fees

ciuffagianluca:
`getFees` to call payment-methods v2 `calculateFees`

### List of changes

- change base policy set backend service for calculateFees 
- edit request body manipulation to match payment-methods calculateFees V2

api test in uat after apply [here](https://dev.azure.com/pagopaspa/pagoPA-projects/_build/results?buildId=219074&view=results)

Screenshot 2025-06-20 alle 11 01 23

### Motivation and context

Handling getFees v2 for app IO calculateFees

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [x] No

### Other information

---


sfarralorenzo (approved):
🚀 
Small note: checking the openAPI fields for both v1 and v2, the only missing field I can see is `primaryCiIncurredFee`. It's not mandatory and it's indeed removed from v2, so it's probably what we want, right?

infantesimone (approved):

PR #3171, opened 2025-06-23
feat: [CHK-4246] - inject transactions service api key into policies

pietro-tota:
Depends on https://github.com/pagopa/pagopa-infra/pull/3169

### List of changes
Set x-api-key into all policies that invoke ecommerce-transactions-service api's

### Motivation and context
Those modification are needed in order to properly inject api key into all service api calls

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information
applied with following targets
```code
sh terraform.sh apply weu-dev \
--target=azurerm_api_management_named_value.ecommerce_transactions_service_api_key_value \
--target=azurerm_api_management_api_operation_policy.create_transactions_v2 \
--target=azurerm_api_management_api_operation_policy.npg_notifications_policy \
--target=module.apim_ecommerce_checkout_api_v1 \
--target=module.apim_ecommerce_checkout_api_v2 \
--target=module.apim_ecommerce_checkout_api_v3 \
--target=module.apim_ecommerce_io_api_v2 \
--target=module.apim_ecommerce_redirect_outcome_api_v1 \
--target=module.apim_ecommerce_transaction_auth_requests_service_api_v1 \
--target=module.apim_ecommerce_transaction_auth_requests_service_api_v2 \
--target=module.apim_ecommerce_transaction_user_receipts_service_api_v1 \
--target=module.apim_ecommerce_transactions_service_api_v1 \
--target=module.apim_ecommerce_transactions_service_api_v2 \
--target=module.apim_ecommerce_webview_api_v1 \
--target=module.apim_ecommerce_webview_api_v2


//nodo app 
sh terraform.sh apply weu-dev \
--target=module.apim_api_receipt_for_ndp_api_v1

//core 
sh terraform.sh apply dev \
--target=azurerm_api_management_api_operation_policy.send_payment_result_api_v2_wisp_policy
```

---


src/domains/ecommerce-app/04_apim_ecommerce.tf
```diff
@@ -741,3 +741,22 @@ resource "azurerm_api_management_named_value" "ecommerce_payment_requests_api_ke
...
+  name         = "ecommerce-transactions-service-primary-api-key"
+  key_vault_id = data.azurerm_key_vault.kv.id
+}
+
+data "azurerm_key_vault_secret" "ecommerce_transactions_service_secondary_api_key" {
+  name         = "ecommerce-transactions-service-secondary-api-key"
+  key_vault_id = data.azurerm_key_vault.kv.id
+}
+
+resource "azurerm_api_management_named_value" "ecommerce_transactions_service_api_key_value" {
```
CianoDanilo, line 755: For the payment-request, we created a different named value for every api using the api key (see [here](https://github.com/pagopa/pagopa-infra/pull/3127/files)). What do you think about this approach here?

pietro-tota: Yes, I see but:
- a gradual approach can lead to misalignment: potentially we can leave some api with an old key version that after applicative rotation can stop working, breaking api partially.
- In case of "per-client api key" we should perform refactoring also in application logic, leaving a single named variable here could potentially give us a way to perform a zero downtime deploy playing with n.v. used for single api vs service accepted api keys.

My concern here is that those named value are not domain segregated and duplicating they is effectively a code duplication: can lead to maintainability issued and to number explosion where the same named value is duplicated per api and per service with no added value

CianoDanilo: It is a duplication of code but not every duplication of code is bad practice. Regarding the first point, we could avoid misalignment, we have application logs that we can enable to check which api key the client is using. I believe there is added value in not breaking all the APIs at the same time and in having the opportunity to observe any issues on a per-API basis.

sfarralorenzo (dismissed):

src/domains/ecommerce-app/04_apim_ecommerce.tf
dylantangredi-jakala: So obviously both approaches are valid and have pros and cons/trade-offs, but for my 2 cents I would be leaning a bit towards the "single/same named values" adopted in this PR. The reason is because I am giving lots of importance to the "partial failures" that could be happening with the separate named values, because imho the partial failure could be more dangerous than a complete one, in the sense that if something goes wrong in the key rotation the system could appear to be "mostly working" and the problem might not be detected immediately, while choosing the same named value when it makes sense the failure would be imemdiately obvious with instant alerts across the service. 

In the end, as Pietro mentioned, we already support primary/secondary key switching as a "safety" mechanism, so the benefits of the gradual rollout could be redundant. If in the future, like Danilo mentioned, we needed to use an api key client it will be relatively simple to change the named values to single ones, but as of now it would be solving a problem we do not have yet.

Good discussion with very good insights guys 👏🏻

dylantangredi-jakala (dismissed):

infantesimone (approved):

PR #3172, opened 2025-06-23
feat(api-key): [CHK-4251] add payment-methods service API keys

CianoDanilo:
### List of changes

Added payment-methods service API keys
### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/pay-wallet-common/02_security.tf
```diff
@@ -329,6 +329,18 @@ resource "azurerm_key_vault_secret" "payment_wallet_gha_bot_pat" {
   }
 }
 
+resource "azurerm_key_vault_secret" "ecommerce_payment_methods_active_api_key" {
```
pietro-tota, line 332: Using a un-linked secret for payment wallet domain could cause to forgot this secret to be updated manually and then misalign environment. Moreover this secret is actually updated manually only, so a misalignment could not be detected by a plan.
I found it a bit risky in case a key rotation is performed, let's say, 2 years from now (could be forgotten), and having half key rotation process automatic and half manual is not so good.

We can solve this with two approach:
1. use a single named value for all api, defined and handled in eCommerce domain
2. fetching secret value from eCommerce through a data and use it to populate this secret

it's generally speaking ok leave two secret for two domains but since they have to be synchronized (as per current service implementation) I would not leave one to manual modification.

I personally prefer the first approach (you know from the previous PR 😄) where the api key rotation is automatically and atomically propagated to every usage point

dylantangredi-jakala (approved):

PR #3245, opened 2025-07-17
fix: fix policy to include readers

FedericoRuzzier:
### List of changes
- change display name of jwt apis
- change policy for marketplace apis in apiconfig fe

### Motivation and context
Needed in order to create payment methods

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/afm-app/api/marketplace-service/jwt_v1/_base_policy.xml
```diff
@@ -40,9 +40,6 @@
           <claim name="aud">
             <value>${apiconfig_fe_client_id}</value>
           </claim>
-          <claim name="groups" match="any">
```
pasqualespica, line 43: why remove "pagopa-ENV-apiconfig-writers" claim ?

PR #3261, opened 2025-07-23
feat(alert): [PIDM-835] tweak payment_wallet_npg_notifications_availability_v1

CianoDanilo:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/pay-wallet-common/00_alerts.tf
```diff
@@ -199,7 +199,7 @@ AzureDiagnostics
 | summarize
     Total=count(),
     Success=countif(responseCode_d < 500 and DurationMs < 350)
-    by Time = bin(TimeGenerated, 15m)
+    by Time = bin(TimeGenerated, 10m)
```
ciuffagianluca, line 202: Why set 10 minutes for timebin? Didn't we say to trying adjust only traffic threshold for this step?

infantesimone (approved):

PR #3279, opened 2025-07-30
feat: cosmosdb alert for provisioned throughput

pietro-tota:
### List of changes
Tune currently deployed alerts performing some tuning:
- change region filter using mapping -> the actually used one (no spaces, all lower cased) is wrong and does not select any value during query execution
- change alert metric from Normalized consumption to provisioned throughput -> normalized consumption can be 100% without being a real problem, it just means that all the provisioned throughput is used in a given period. Alert must be triggered once reaching 100% of provisioned throughput since it means that DB have reached it's maximum capacity and cannot any scale up
- change alert window parameter -> having same time window and frequency can lead to some data being lost due to datapoint that have not being collected yet. Changed this value setting a time windows greater than alert evaluation frequency

### Motivation and context
Those modifications are needed to fix CosmosDB utilization alert query in order to be alerted promptly 

### Type of changes

- [x] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information
Tested in dev environment creating alert (without opsgenie action) and using prod db as datasource for the query, checking that datapoint are evaluated correctly

---


infantesimone (approved):

infantesimone:
well done!

PR #3310, opened 2025-09-02
fix: handle error parsing response body

pietro-tota:
### List of changes
Handle errors parsing response body when it's not a valid json

### Motivation and context
This modification will handle errors parsing response body when it's not a valid json returning response unaltered

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information
Applied with following target
```sh 
sh terraform.sh apply weu-dev \                                                                                                                                                           ─╯
--target=module.apim_api_bizevents_lap_api_jwt_v1 \
--target=module.apim_api_bizevents_lap_api_v1
```

---


pasqualespica (approved):
lgtm

PR #3358, opened 2025-09-18
feat: [CHK-4451] add payment methods handler api

ciuffagianluca:
Payment methods handler openapi

### List of changes

Add payment methods handler openapi definition

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---

### DIFF
--- a/src/domains/shared-app/.terraform.lock.hcl
+++ b/src/domains/shared-app/.terraform.lock.hcl
@@ -61,6 +61,25 @@ provider "registry.terraform.io/hashicorp/azurerm" {
   ]
 }
 
+provider "registry.terraform.io/hashicorp/external" {
+  version = "2.3.5"
+  hashes = [
+    "h1:FnUk98MI5nOh3VJ16cHf8mchQLewLfN1qZG/MqNgPrI=",
+    "zh:6e89509d056091266532fa64de8c06950010498adf9070bf6ff85bc485a82562",
+    "zh:78d5eefdd9e494defcb3c68d282b8f96630502cac21d1ea161f53cfe9bb483b3",
+    "zh:86868aec05b58dc0aa1904646a2c26b9367d69b890c9ad70c33c0d3aa7b1485a",
+    "zh:a2ce38fda83a62fa5fb5a70e6ca8453b168575feb3459fa39803f6f40bd42154",
+    "zh:a6c72798f4a9a36d1d1433c0372006cc9b904e8cfd60a2ae03ac5b7d2abd2398",
+    "zh:a8a3141d2fc71c86bf7f3c13b0b3be8a1b0f0144a47572a15af4dfafc051e28a",
+    "zh:aa20a1242eb97445ad26ebcfb9babf2cd675bdb81cac5f989268ebefa4ef278c",
+    "zh:b58a22445fb8804e933dcf835ab06c29a0f33148dce61316814783ee7f4e4332",
+    "zh:cb5626a661ee761e0576defb2a2d75230a3244799d380864f3089c66e99d0dcc",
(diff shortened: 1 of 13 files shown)


src/domains/ecommerce-app/04_apim_ecommerce.tf
```diff
@@ -856,3 +926,22 @@ resource "azurerm_api_management_named_value" "ecommerce_helpdesk_command_servic
...
+}
+
+resource "azurerm_api_management_named_value" "ecommerce_payment_methods_handler_api_key_value" {
+  name                = "ecommerce-payment-methods-handler-api-key-value"
+  api_management_name = local.pagopa_apim_name
+  resource_group_name = local.pagopa_apim_rg
+  display_name        = "ecommerce-payment-methods-handler-api-key-value"
+  value               = var.ecommerce_payment_methods_handler_api_key_use_primary ? data.azurerm_key_vault_secret.ecommerce_payment_methods_handler_primary_api_key.value : data.azurerm_key_vault_secret.ecommerce_payment_methods_handler_secondary_api_key.value
+  secret              = true
+}
```
CianoDanilo, line 947: We chose to use payment-methods API keys, we can avoid creating and using new API keys for this service

ciuffagianluca: Why using same api key for two different services? We'll continue to use payment methods for NPG proxying or not? In this case we need 2 different api key, don't you think?

CianoDanilo: We will not, the final goal (not of this sprint) is to port every API that is currently handled by the payment-methods-service to payment-methods-handler, resulting in the retirement of that service. This is why we chose to keep using the same API key and to momentarily share it between the two services.
If we want to change what we agreed on before, please let me know and I will update the other PRs in accordance with the decision.
@infantesimone @pietro-tota @dylantangredi-jakala @forciniti-luigi-jakala

ciuffagianluca: Ok, I missed the migration of all other api. I'll remove this resources

pietro-tota: Yes, we agreed to go with the same api key since this will be a transitory where the payment methods service will be dismantled

pietro-tota (dismissed):
LGMT, just take care of the @CianoDanilo comment

src/domains/ecommerce-app/api/ecommerce-payment-methods-handler/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,455 @@
...
+    "security": [
+        {
+            "ApiKeyAuth": []
+        }
+    ],
+    "paths": {
+        "/payment-methods": {
+            "post": {
+                "tags": [
+                    "payment-methods"
```
pietro-tota, line 39:
```suggestion
                    "payment-methods-handler"
```

src/domains/ecommerce-app/api/ecommerce-payment-methods-handler/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,455 @@
...
+                ],
+                "operationId": "getAllPaymentMethods",
+                "summary": "Retrieve all Payment Methods (by filter)",
+                "description": "API for retrieve payment method using the request query parameter filters",
```
pietro-tota, line 43:
```suggestion
                "description": "GET with request body, no resource will be created: API for retrieve payment method using the request query parameter filters",
```

pietro-tota (approved):

PR #3738, opened 2026-03-12
feat: Add Debizium filterArchived

cap-ang:
### List of changes

- Update Debizium kafka-connect.yaml, new version v2 with groovy jar and Debizium scripting.
- Add filterArchived transforms in Debizium KafkaConnector PostgresConnector to filter events related to archived debt positions.

### Motivation and context

- Exclude archived events from CDC

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


src/domains/gps-app/yaml/postgres-connector.yaml
```diff
@@ -24,4 +24,13 @@ spec:
...
+
+    transforms: "filterArchived"
+    transforms.filterArchived.type: "io.debezium.transforms.Filter"
+    transforms.filterArchived.language: "jsr223.groovy"
+
+    # Groovy filter logic:
+    # 1. If value is null (Tombstone): keep
+    # 2. If it is a DELETE (d): keep ONLY if before.archived is false
+    # 3. If it is INSERT/UPDATE: keep ONLY if after.archived is false
+    transforms.filterArchived.condition: "value == null ? true : (value.op == 'd' ? value.before?.archived != true : value.after?.archived != true)"
```
gioelemella, line 36: if i am not mistaken, this condition can not work with the current CDC configuration, on delete events only the id is present https://debezium.io/documentation//reference/3.0/connectors/postgresql.html#postgresql-delete-events

src/domains/gps-app/yaml/postgres-connector.yaml
```diff
@@ -24,4 +24,13 @@ spec:
...
     snapshot.max.threads: ${max_threads}
     publication.autocreate.mode: "disabled" # shall be create before
-    # publication.autocreate.mode: "filtered"  # create it
+
+    transforms: "filterArchived"
+    transforms.filterArchived.type: "io.debezium.transforms.Filter"
+    transforms.filterArchived.language: "jsr223.groovy"
+
+    # Groovy filter logic:
+    # 1. If value is null (Tombstone): keep
```
svariant, line 33: Why keep the tombstone events?

cap-ang: EventHub needs tombstone event after delete.

cap-ang: _Debezium emits these when a row is deleted in the source database to signal to Kafka topics (with log compaction enabled) that all prior records with that key should be permanently deleted._

cap-ang: For Debezium, Kafka uses Log Compaction. A compacted topic doesn't delete data based on time (eventhub retention).
Kafka logs are immutable. To delete a record, we have to append a new message that acts as a deletion marker: the tombstone event.

So, to save space, we need to keep them.

src/domains/gps-app/yaml/postgres-connector.yaml
cap-ang: That's correct. This field will be added by updating the identity replica.

jacopocarlini (approved):

FedericoRuzzier (approved):

PR #3794, opened 2026-04-16
CHORE: [PIDM-1805] Updated alert threshold to require two consecutive bins to

alessio-acitelli:
### List of changes

- Updated the Terraform configuration of the pagopa-p-nodo-nodo_for_pa_auth-api-availability Azure Monitor scheduled query alert.
- Changed the alert query logic to trigger only when two consecutive 5-minute bins are below the configured availability threshold, instead of firing on a single degraded bin.

### Motivation and context

This change reduces alert sensitivity for the pagopa-fdr-nodo-rest-availability monitor by filtering out short-lived transient degradations.

With this update, the alert is triggered only when the degradation persists for two consecutive 5-minute intervals, making it less sensitive and more aligned with sustained service issues.

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


FedericoRuzzier (approved):
LGTM 👍

pasqualespica (approved):

PR #3878, opened 2026-06-08
chore: Enable azure data factory schedule

jacopocarlini:
### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/gps-common/04_datafactory_pipeline.tf
```diff
@@ -182,7 +182,7 @@ resource "azapi_resource" "pipeline_lifecycle_management" {
 }
 
 resource "azurerm_data_factory_trigger_schedule" "pipeline_lifecycle_management_schedule" {
```
cap-ang, line 184: Could you add `start_time` parameter?

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_factory_trigger_schedule#start_time-1

cap-ang (changes requested):

src/domains/gps-common/04_datafactory_pipeline.tf
jacopocarlini: done

PR #3908, opened 2026-06-17
chore: [CHK-4874] removed jwt validation

mcapponi-jakala:
### List of changes

- Symmetric key validation of JWT tokens has been removed from APIM policies. The token is now validated using only the jwt-issuer-service.

### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


CianoDanilo (approved):
LGTM, I would just split this PR in multiple smaller PRs by domain/impacted APIs

sterondinella (approved):

PR #3938, opened 2026-07-03
chore: remove responder from SNAT connections alert

pietro-tota:
### List of changes
Remove opsgenie team core responder from SNAT over 10k connections alert

### Motivation and context
This alert is already routed to pagopa-infra team, no need to route it to dev team too

### Type of changes

- [ ] Add new resources

- [x] No

- [ ] Yes
- [x] No

### Other information

---


pasqualespica (approved):
lgtm

PR #3939, opened 2026-07-03
[PIDM-2149] selfcare  multiple role management

FrancescoDiTonnoJakala:
### List of changes
New APIM policy with an exchange token call to handle:
- mapping of all roles received from AR within the 'roles' claim
- addition of a query parameter for role selection (optional)
- default role selection if no choice is made (selection query parameter = null)
- path structure aligning with best practices used in other domains—following the `/domain/version/api` format—by creating a new API group with the path `/backoffice/selfcare/sso/v1/tokens`, where the logic of the previously created API is replicated and modified to incorporate these changes
- Addition of a feature flag to enable the new API from the frontend

### Type of changes

- [x] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information

---


src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,134 @@
...
+                    Jwt selcToken = (Jwt)context.Variables["outputToken"];
+                    string organizationClaim = selcToken.Claims.GetValueOrDefault("organization", "");
+
+                    if (string.IsNullOrWhiteSpace(organizationClaim))
+                    {
+                        return "";
+                    }
+
+                    JObject organization = JObject.Parse(organizationClaim);
+                    var org_id = organization.Value&lt;string>("id");
```
pietro-tota:
```suggestion
                    var org_id = organization.Value<string>("id");
```

please fix all other escaped '<' characters

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,134 @@
...
+                    return message;
+                }' />
+        <choose>
+            <when condition='@(string.IsNullOrEmpty((string)context.Variables["pagopaPortalToken"]))'>
+                <return-response>
+                    <set-status code="401" reason="Unauthorized" />
+                    <set-header name="Content-Type" exists-action="override">
+                        <value>text/plain</value>
+                    </set-header>
+                    <set-body>The desired role is missing or is not assigned to the user.</set-body>
```
pietro-tota:
```suggestion
                    <set-body>Unauthorized</set-body>
```

let make unauthorized error message as generic as possible -> the scope here is to hide errors due to user input to not allow an attacker to discover implementation details due to error responses

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,134 @@
...
+                <issuer>${selfcare-issuer}</issuer>
+            </issuers>
+        </validate-jwt>
+        <set-variable name="pagopaPortalToken" value='@{
+                    Jwt selcToken = (Jwt)context.Variables["outputToken"];
+                    string organizationClaim = selcToken.Claims.GetValueOrDefault("organization", "");
+
+                    if (string.IsNullOrWhiteSpace(organizationClaim))
+                    {
+                        return "";
```
pietro-tota: instead of returning an empty string here we can throw an exception with a custom error so that it can make clearer diagnostic error cases tracing api call with azure portal

something like 
throw new Exception("organization claim cannot be empty");

etc to all fields

then add an onError handling policy like done [here](https://github.com/pagopa/pagopa-infra/blob/f4de281efcabae00d9bdcb9efce9a9fa8021e138/src/domains/ecommerce-app/api/npg-notification/_npg_notifications_policy.xml.tpl#L179)
to trace errors

src/domains/selfcare-common/04_app_configuration_feature.tf
```diff
@@ -97,6 +97,21 @@ resource "azurerm_app_configuration_feature" "commission_bundles_flag" {
   }
 }
 
+resource "azurerm_app_configuration_feature" "profile_item_flag" {
+  configuration_store_id = azurerm_app_configuration.selfcare_appconf.id
+  description            = "It enables the profile item role switch"
+  name                   = "profile-item"
+  enabled                = true
```
pietro-tota:
```suggestion
  enabled                = false
```

we should disable ff initially and enable it after UAT integration tests.
So we can disable it now and enable in each env as needed

FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/510145b842b6c972e89b294c1ecbf4bfa1ffe840

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/510145b842b6c972e89b294c1ecbf4bfa1ffe840

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/38601d7f9065847199976973e1db0a34c66d92aa

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-infra/pull/3939/changes/38601d7f9065847199976973e1db0a34c66d92aa

pietro-tota (dismissed):

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,64 @@
...
+            "name": "IdentityToken",
+            "in": "header",
+            "description": "Selfcare identity token.",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "desidered_role",
```
dylantangredi-jakala: typo, better correct it now that the api is new (also change the reference on the policy please)

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,139 @@
...
+            <metadata name="errorSource" value="@(context.LastError.Source)" />
+            <metadata name="errorMessage" value="@(context.LastError.Message)" />
+            <metadata name="errorReason" value="@(context.LastError?.Reason ?? "-")" />
+            <metadata name="errorSection" value="@(context.LastError?.Section ?? "-")" />
+            <metadata name="errorPath" value="@(context.LastError?.Path ?? "-")" />
+            <metadata name="errorStatusCode" value="@((context.Response?.StatusCode ?? -1).ToString())" />
+        </trace>
+        <base />
+        <return-response>
+            <set-status code="401" reason="Unauthorized" />
```
dylantangredi-jakala: we tested this together but can you confirm that every error (including internal 500s like a malformed organization claim or a missing signing cert) has to be returned as 401 Unauthorized?

src/domains/selfcare-app/04_apim_token_exchange.tf
```diff
@@ -169,3 +169,55 @@ resource "azurerm_api_management_api_operation_policy" "pagopa_token_exchange_po
...
+    display_name          = "pagoPA Backoffice Selfcare SSO API"
+    description           = "Exchange a Selfcare token to a pagoPA platform token"
+    path                  = "backoffice/selfcare/sso"
+    subscription_required = false
+    service_url           = null
+  }
+}
+
+resource "azurerm_api_management_api_version_set" "pagopa_backoffice_selfcare_sso_v1" {
+  name                = "${var.env_short}-pagopa-backoffice-selfcare-sso-v1"
```
dylantangredi-jakala: a version set groups all versions, so -v1 doesn't belong in its name => a future v2 would live in a set named …-sso-v1. Existing sets (`apim_cache_v1`, `node_for_ecommerce_api`) omit the version from the set name. 
Safe to rename now since the resource has no state yet. e.g.


```suggestion
  name                = "${var.env_short}-pagopa-backoffice-selfcare-sso"
```

dylantangredi-jakala (commented):
LGTM, only minor issues

src/domains/selfcare-app/04_apim_token_exchange.tf
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-infra/pull/3939/changes/d936c7793bff3ec3efae5a3fcb87706c7015622b

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_openapi.json.tpl
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-infra/pull/3939/changes/d936c7793bff3ec3efae5a3fcb87706c7015622b

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Added 500 error in addition to 401 here https://github.com/pagopa/pagopa-infra/pull/3939/changes/d936c7793bff3ec3efae5a3fcb87706c7015622b

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,152 @@
...
+                    if (roles == null || roles.Count == 0)
+                    {
+                        throw new Exception("pagopaPortalToken: organization.roles cannot be empty");
+                    }
+
+                    
+                    string desiredRole = context.Request.Url.Query.GetValueOrDefault("desired_role", "").Trim();
+                    JObject selectedRole = string.IsNullOrWhiteSpace(desiredRole)
+                        ? roles.OfType<JObject>().FirstOrDefault()
+                        : roles.OfType<JObject>().FirstOrDefault(role => role.Value<string>("role") == desiredRole);
```
dylantangredi-jakala: The default is `roles[0]`, but the ordering of the roles array in the Selfcare token is not a documented contract, so a multi-role user can get a different default between logins? if that is the case could we make the default deterministic (highest privilege, or an explicit ordering) rather than positional?

dylantangredi-jakala: The comparison is exact and case sensitive, so a `desired_role=ADMIN` from the frontend returns 401. Should we use a safer comparison instead (unless selfcare guarantees the casing)? e.g. 


```suggestion
                    JObject selectedRole = string.IsNullOrWhiteSpace(desiredRole)
                        ? roles.OfType<JObject>().FirstOrDefault()
                        : roles.OfType<JObject>().FirstOrDefault(role => string.Equals(role.Value<string>("role"), desiredRole, StringComparison.OrdinalIgnoreCase));
```

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,152 @@
...
+                <origin>https://${origin}</origin>
+            </allowed-origins>
+            <allowed-methods>
+                <method>POST</method>
+                <method>OPTIONS</method>
+            </allowed-methods>
+            <allowed-headers>
+                <header>*</header>
+            </allowed-headers>
+        </cors>
```
dylantangredi-jakala, line 14: missing `<base />`, intentional?

src/domains/selfcare-app/04_apim_token_exchange.tf
```diff
@@ -169,3 +169,55 @@ resource "azurerm_api_management_api_operation_policy" "pagopa_token_exchange_po
...
+  pagopa_backoffice_selfcare_sso_v1_api = {
+    display_name          = "pagoPA Backoffice Selfcare SSO API"
+    description           = "Exchange a Selfcare token to a pagoPA platform token"
+    path                  = "backoffice/selfcare/sso"
+    subscription_required = false
+    service_url           = null
+  }
+}
+
+resource "azurerm_api_management_api_version_set" "pagopa_backoffice_selfcare_sso_v1" {
```
dylantangredi-jakala: the name attribute is fixed, but the resource label still carries `_v1`, so a future `v2` would be attached to a set labelled `..._sso_v1`, intended?

src/domains/selfcare-app/04_apim_token_exchange.tf
```diff
@@ -169,3 +169,55 @@ resource "azurerm_api_management_api_operation_policy" "pagopa_token_exchange_po
...
+  resource_group_name = local.pagopa_apim_rg
+  api_management_name = local.pagopa_apim_name
+  display_name        = local.pagopa_backoffice_selfcare_sso_v1_api.display_name
+  versioning_scheme   = "Segment"
+}
+
+module "apim_pagopa_backoffice_selfcare_sso_v1" {
+  source = "./.terraform/modules/__v4__/api_management_api"
+
+  name                  = "${local.project}-backoffice-selfcare-sso"
```
dylantangredi-jakala: `local.project` already ends in -selfcare, right? if so, this resolves to `pagopa-<env>-weu-selfcare-backoffice-selfcare-sso-v1` with "selfcare" twice, should we trim it?

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_openapi.json.tpl
```diff
@@ -0,0 +1,64 @@
...
+          "401": {
+            "description": "The Selfcare identity token is invalid or the requested role is not assigned to the user.",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "type": "string"
+                }
+              }
+            }
+          }
```
dylantangredi-jakala, line 79: we are mixing "identity token invalid" and "role not assigned", which are different recoveries for the client (re-login vs pick another role), intended?

Also worth documenting the 500 the policy can return, since the otherwise branch in the `on-error` section produces it, what do you think?

dylantangredi-jakala (commented):
also Validate PR title is failing, i think the action needs the pr name to have the feat/fix etc keyword as prefix

also worth rerunning the pre-commit check

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: There is no default option at the moment, a change with a sorting would lead to an increase in the complexity of the policy, the possibility of a future modification is not excluded

dylantangredi-jakala: fair enough

FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/1a6e43232dd64fab48e69933004521bc0f87b2b6

src/domains/selfcare-app/04_apim_token_exchange.tf
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/1a6e43232dd64fab48e69933004521bc0f87b2b6

src/domains/selfcare-app/04_apim_token_exchange.tf
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/1a6e43232dd64fab48e69933004521bc0f87b2b6

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/1a6e43232dd64fab48e69933004521bc0f87b2b6

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_openapi.json.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/1a6e43232dd64fab48e69933004521bc0f87b2b6

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
dylantangredi-jakala: any specific reason why we added it before `<cors>`? I think every other policy adds it after, e.g. 
https://github.com/pagopa/pagopa-infra/blob/04d65c01193b8b8ac5bc47f2c0ddfe493c89b00f/src/domains/checkout-app/api/checkout/checkout_feature_flags/v1/_base_policy.xml.tpl#L3-L15

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,162 @@
...
+                    string organizationClaim = selcToken.Claims.GetValueOrDefault("organization", "");
+
+                    if (string.IsNullOrWhiteSpace(organizationClaim))
+                    {
+                        throw new Exception("pagopaPortalToken: organization claim cannot be empty");
+                    }
+
+                    JObject organization = JObject.Parse(organizationClaim);
+                    var org_id = organization.Value<string>("id");
+                    var org_vat = organization.Value<string>("fiscal_code");
```
dylantangredi-jakala, line 36: any reason why id and fiscal_code are not guarded like the other fields? dont we risk generating a valid token with a null id instead of throwing a 401? something like 

```
var org_id = organization.Value<string>("id");
var org_vat = organization.Value<string>("fiscal_code");

if (string.IsNullOrWhiteSpace(org_id) || string.IsNullOrWhiteSpace(org_vat)) {
    throw new Exception("pagopaPortalToken: organization id or fiscal_code cannot be empty");
}
```

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,162 @@
...
+            <message>pagoPA Backoffice Selfcare SSO Error</message>
+            <metadata name="errorSource" value="@(context.LastError.Source)" />
+            <metadata name="errorMessage" value="@(context.LastError.Message)" />
+            <metadata name="errorReason" value="@(context.LastError?.Reason ?? "-")" />
+            <metadata name="errorSection" value="@(context.LastError?.Section ?? "-")" />
+            <metadata name="errorPath" value="@(context.LastError?.Path ?? "-")" />
+            <metadata name="errorStatusCode" value="@((context.Response?.StatusCode ?? -1).ToString())" />
+        </trace>
+        <base />
+        <choose>
```
dylantangredi-jakala, line 138: is `<base />`  before `<choose>` correct here? `<base />` inherits the global on-error scope, so if that scope ever returns a response, this `<choose>` never runs and the 401/403 mapping is lost maybe?

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
```diff
@@ -0,0 +1,162 @@
...
+                <origin>https://${origin}</origin>
+            </allowed-origins>
+            <allowed-methods>
+                <method>POST</method>
+                <method>OPTIONS</method>
+            </allowed-methods>
+            <allowed-headers>
+                <header>*</header>
+            </allowed-headers>
+        </cors>
```
dylantangredi-jakala, line 14: non-blocking, do you think we need a rate limiting here since each call costs a jwt validation plus an RSA sign before any rejection? e.g. 

https://github.com/pagopa/pagopa-infra/blob/04d65c01193b8b8ac5bc47f2c0ddfe493c89b00f/src/domains/checkout-app/api/checkout/checkout_feature_flags/v1/_base_policy.xml.tpl#L16

FrancescoDiTonnoJakala: Good point, but I'd skip it for now: the RSA sign only runs after a successful validate-jwt, so a caller still needs a valid SelfCare IdentityToken to reach the expensive part. Given that, plus the low expected volume, I don't think it's worth adding here.

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/f82954ed3e2a4876bf9f2d303e89f5be1670c2d2

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/f82954ed3e2a4876bf9f2d303e89f5be1670c2d2

src/domains/selfcare-app/api/pagopa-selfcare-ms-backoffice/sso/v1/_base_policy.xml.tpl
FrancescoDiTonnoJakala: Done here https://github.com/pagopa/pagopa-infra/pull/3939/changes/f82954ed3e2a4876bf9f2d303e89f5be1670c2d2

jacopocarlini (approved):
💪 great

dylantangredi-jakala (approved):

PR #3986, opened 2026-07-30
fix: role identity

pietro-tota:
### Type of changes

- [ ] Add new resources

- [ ] No

- [ ] Yes
- [ ] No

### Other information
```sh
sh terraform.sh apply weu-xxx \
--target=azurerm_kusto_database_principal_assignment.nodo_pod_identity
```

---


pasqualespica (approved):
lgtm @pietro-tota 
https://github.com/pagopa/pagopa-stand-in-manager and other service can use inodo_workload_identity_kusto_principal_assignment

@aomegax  👀 ❓ into stanngmng i saw that https://github.com/pagopa/pagopa-stand-in-manager/blob/83c8848e8d1f9ddb94e4a855f257925d66db48ff/helm/values-prod.yaml#L38 and this one too https://github.com/pagopa/pagopa-stand-in-manager/blob/83c8848e8d1f9ddb94e4a855f257925d66db48ff/helm/values-prod.yaml#L117

